PR #26, opened 2026-08-22
fix(openai): enforce one-to-one tool-call history identities

### LINKED ISSUE #15: [openai] duplicate tool-call IDs collapse one-to-one history validation
## Summary

The shared OpenAI tool-history validator records expected tool calls in `map[string]bool`. Duplicate non-empty IDs therefore overwrite one another. A turn containing two assistant tool calls with the same ID and only one tool result passes validation, even though the function's contract says every call must be answered exactly once.

Both Chat Completions and Responses use this validator. Chat parsing also preserves duplicate provider-supplied non-empty IDs, and Responses request serialization writes duplicate `call_id` values unchanged.

Relevant code:

- [`validateOpenAIToolHistory` collapses expected calls into `map[string]bool` and never checks duplicate calls/results](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/tool_history.go#L10-L50)
- [Chat request construction relies on this validator](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/chat.go#L1027-L1032)
- [`ensureChatToolCallIDs` only fills empty IDs; duplicate non-empty IDs are retained](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/chat.go#L1414-L1451)
- [Responses request construction uses the same validator](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/responses.go#L1367-L1373)
- [Responses emits each assistant tool call with its original `call_id`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/responses.go#L1421-L1444)

## Deterministic reproduction

```go
messages := []llm.Message{
    {
        Role: llm.RoleAssistant,
        ToolCalls: []llm.ToolCall{
            {ID: "call_dup", Type: "function", Function: llm.FunctionCall{Name: "first", Arguments: `{}`}},
            {ID: "call_dup", Type: "function", Function: llm.FunctionCall{Name: "second", Arguments: `{}`}},
        },
    },
    {
        Role:       llm.RoleTool,
        ToolCallID: "call_dup",
        Content:    llm.TextContent("only one result"),
    },
}

if err := validateOpenAIToolHistory(messages, "openai"); err != nil {
    t.Fatalf("unexpected validation error: %v", err)
}
```

The first map assignment creates `expected["call_dup"] = false`; the second overwrites the same entry. The single result then sets that one entry to `true`, so validation succeeds.

The inverse is also accepted: one assistant call followed by two tool messages with the same `tool_call_id`; both merely assign `true` to the same map entry.

## Actual behavior

Duplicate tool-call identities are accepted, and the validator cannot prove a one-to-one call/result relationship.

## Expected behavior

Assistant tool-call IDs must be unique within a turn, and each ID must have exactly one contiguous result. Duplicate calls or duplicate results must fail locally with an actionable error.

## Impact

- One result may be treated as satisfying multiple different function calls.
- Duplicate provider IDs can enter Agent history and later produce invalid or ambiguous requests.
- Chat and Responses can diverge from actual tool execution state during retry/resume.
- The fail-closed history check reports success for a structurally invalid turn.

This is a direct property of Go map key overwrite semantics and is reproducible without a provider or network. The validator's own comment/invariant requires one result per call, while the data structure stores only one boolean per ID.

## Suggested fix

Track counts and reject duplicates explicitly:

1. While reading assistant calls, reject an ID already present.
2. While reading tool results, reject a second result for an already-satisfied ID.
3. Require every unique call to have count exactly one.
4. Validate provider-parsed tool calls before returning a completion, so duplicate IDs do not enter history.

Add regression tests for duplicate assistant IDs, duplicate results, normal parallel calls, and provider responses containing duplicate IDs.

timwhitez:
## Summary

- reject duplicate non-empty assistant tool-call IDs before request serialization
- reject a second contiguous tool result for an already-satisfied ID
- preserve valid parallel calls whose results arrive in any order

## Why

The previous boolean map silently collapsed duplicate IDs, allowing one result to satisfy multiple calls or multiple results to satisfy one call.

## Validation

- [x] focused duplicate-call, duplicate-result, and valid-parallel regressions
- [x] targeted OpenAI package tests
- [x] temporary verification workflow removed
- [x] final clean diff review completed

Closes #15


timwhitez (commented):
Second-pass review completed after removing temporary CI. The validator now rejects duplicate assistant IDs before result matching and rejects a second result for an already-satisfied ID, while retaining legal out-of-order results for distinct parallel calls. Focused and full repository tests passed on the source commit; the final diff contains only implementation and regression tests. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `8e125994ceed82dca032806dd6c072e8a51571f1`.

- `go build ./...` passed

PR #27, opened 2026-08-22
fix(anthropic): preserve unique tool-call identities after normalization

### LINKED ISSUE #14: [anthropic] lossy tool-call ID normalization allows collisions and invalid history to pass validation
## Summary

Anthropic tool-call IDs are normalized by trimming them and replacing every character outside `[A-Za-z0-9_-]` with `_`. This transformation is not injective: distinct IDs such as `call/a` and `call:a` both become `call_a`.

`validateAnthropicToolHistory` stores normalized IDs in `map[string]bool` without detecting duplicate keys. A collision therefore collapses multiple assistant `tool_use` blocks into one logical entry. One tool result can mark both calls as satisfied, duplicate tool results are also accepted, and serialization emits ambiguous duplicate `tool_use.id` / `tool_result.tool_use_id` values.

Relevant code:

- [`normalizeToolCallIDWithWarning` performs the lossy replacement without collision protection](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/anthropic/client.go#L588-L613)
- [`validateAnthropicToolHistory` keys a boolean map by the normalized ID and overwrites duplicate calls](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/anthropic/client.go#L1329-L1374)
- [Assistant tool calls are serialized with the same normalized ID](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/anthropic/client.go#L1410-L1432)
- [Tool results are independently normalized to that same ID](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/anthropic/client.go#L1457-L1473)

## Deterministic reproduction

```go
messages := []llm.Message{
    {
        Role: llm.RoleAssistant,
        ToolCalls: []llm.ToolCall{
            {ID: "call/a", Type: "function", Function: llm.FunctionCall{Name: "first", Arguments: `{}`}},
            {ID: "call:a", Type: "function", Function: llm.FunctionCall{Name: "second", Arguments: `{}`}},
        },
    },
    {
        Role:       llm.RoleTool,
        ToolCallID: "call/a",
        Content:    llm.TextContent("only one result"),
    },
}

// Both original IDs normalize to "call_a". Despite two tool calls and only
// one result, validation succeeds.
if err := validateAnthropicToolHistory(messages); err != nil {
    t.Fatalf("unexpected validation error: %v", err)
}

_, wire, err := serializeMessages(messages)
if err != nil {
    t.Fatal(err)
}
// wire[0] contains two tool_use blocks with id="call_a";
// wire[1] contains one tool_result with tool_use_id="call_a".
```

A second reproduction with two tool messages whose original IDs collide is also accepted because the same boolean is simply assigned `true` twice.

## Actual behavior

The local validator claims the history is valid even though identity was lost and the serialized request cannot establish a one-to-one relationship between tool calls and results.

## Expected behavior

Each assistant tool call must retain a unique wire identity, and each must be matched by exactly one result. Normalization must never silently merge two source IDs.

## Impact

- Valid cross-provider histories can be transformed into invalid Anthropic requests.
- A tool result can be associated with the wrong tool call after normalization.
- The SDK's stated fail-closed history validation is bypassed.
- Retried/resumed multi-tool turns can fail with provider HTTP 400 after local validation reported success.

The collision is purely deterministic string transformation. The map assignment `expected[id] = false` overwrites the first call, and the final loop looks up the same key for both original calls. No provider timing or undocumented behavior is required to demonstrate that the validator's own “exactly one result per tool_use” invariant is violated.

## Suggested fix

Before serializing a turn:

1. Build a mapping from original ID to normalized wire ID.
2. Reject duplicate original IDs and any normalized collision, or generate a collision-resistant legal ID (for example, legal prefix plus a stable hash suffix).
3. Track result counts rather than booleans and require exactly one result for every tool call.
4. Use the same immutable mapping for both `tool_use` and `tool_result` serialization.

Add regression tests for `call/a` vs `call:a`, duplicate original IDs, duplicate results, and a normal multi-tool turn.

timwhitez:
## Summary

- retain source tool-call IDs separately from Anthropic wire-safe IDs
- reject duplicate original IDs and lossy normalization collisions
- reject duplicate tool results
- verify valid source IDs still serialize to matching normalized `tool_use`/`tool_result` IDs

## Validation

- [x] deterministic collision and duplicate-result regressions added
- [x] targeted Anthropic package tests
- [x] final clean diff review completed after temporary verification files were removed

Closes #14


timwhitez (commented):
Second-pass review completed after CI and branch cleanup. I rechecked the source-ID versus wire-ID separation, duplicate-result handling, and serialization regression. The final diff is limited to the Anthropic history validator and focused tests; targeted and full repository test suites passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `dcaf49a3f8c4475a15583c453da728a908559f9d`.

- `go build ./...` passed

PR #28, opened 2026-08-22
fix(sandbox): make apply_patch moves non-clobbering

### LINKED ISSUE #18: [sandbox] apply_patch move can silently overwrite an existing destination on Unix
## Summary

`apply_patch`'s `*** Move to:` implementation validates the source and the destination directory, but never checks whether the destination path already exists. It then calls `os.Rename(fromPath, toPath)` directly.

On Unix, renaming a file onto an existing regular file atomically replaces the destination, silently destroying its contents. On Windows, the same call normally fails when the destination exists. The same patch therefore has destructive success on one supported platform and failure/partial application on another.

Relevant code:

- [`planMove` resolves the destination and validates only its parent directory; it does not reject an existing destination](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_patch.go#L442-L478)
- [`commitAction` creates/probes the directory and passes the two paths directly to `os.Rename`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_patch.go#L570-L592)
- `planAdd` already rejects existing targets, establishing the non-clobbering behavior used for explicit file creation.

## Deterministic reproduction

```go
func TestApplyPatchMoveRejectsExistingDestination(t *testing.T) {
    if runtime.GOOS == "windows" {
        t.Skip("the bug is the Unix overwrite; Windows demonstrates the divergence")
    }

    root := t.TempDir()
    s, err := New(root)
    if err != nil {
        t.Fatal(err)
    }

    if err := os.WriteFile(filepath.Join(root, "source.txt"), []byte("source\n"), 0o644); err != nil {
        t.Fatal(err)
    }
    if err := os.WriteFile(filepath.Join(root, "destination.txt"), []byte("important\n"), 0o644); err != nil {
        t.Fatal(err)
    }

    patch := `*** Begin Patch
*** Update File: source.txt
*** Move to: destination.txt
@@ -1,1 +1,1 @@
-source
+moved
*** End Patch`

    _, err = applyPatchToSandbox(s, patch)
    if err == nil {
        t.Fatal("expected destination-exists error")
    }

    got, _ := os.ReadFile(filepath.Join(root, "destination.txt"))
    if string(got) != "important\n" {
        t.Fatalf("existing destination was clobbered: %q", got)
    }
}
```

Current Unix behavior: the call succeeds, `source.txt` disappears, and `destination.txt` becomes `moved\n`.

Current Windows behavior: `os.Rename` returns an error after the preceding staged write has already modified `source.txt`, so the patch may report partial application.

## Actual behavior

An existing destination is overwritten without an explicit delete/overwrite instruction on Unix; behavior differs on Windows.

## Expected behavior

A move should fail closed when the destination exists unless the patch format exposes an explicit, confirmed overwrite operation. The behavior should be consistent across supported operating systems.

## Impact

- Silent loss of an unrelated destination file.
- Platform-dependent patch results and tests.
- On Windows, partial patch application can leave the source updated but not moved.
- A model typo in `*** Move to:` can destroy a valid workspace file.

The behavior follows documented `os.Rename` platform semantics and requires only two ordinary files. No race or provider behavior is involved. The code has no destination-existence branch and the test suite contains no move-conflict contract.

## Suggested fix

1. During planning, `Lstat` the destination and reject it if it exists (including a path staged by an earlier operation).
2. Record the destination's non-existence and re-check it immediately before commit to prevent races.
3. If overwrite is desired, represent it explicitly and implement a consistent cross-platform replacement with rollback/partial-application handling.
4. Add Unix and Windows tests for existing destination, missing destination, staged destination collisions, and destination creation between planning and commit.

timwhitez:
## Summary

- reject a move destination that already exists on disk or in the staged patch state
- re-check destination non-existence immediately before commit
- atomically create the destination without replacement using a hard link, then remove the source
- roll back the destination link if source removal fails
- fail closed on filesystems that cannot provide the required non-clobbering primitive

## Validation

- [x] pre-existing destination regression
- [x] destination-created-after-planning regression
- [x] destination-created-after-final-check regression
- [x] complete Sandbox tests

Closes #18


timwhitez (commented):
Second-pass review completed after CI and the atomicity correction. The final move primitive creates the destination name without replacement, preserves a last-moment competing file, removes the source only after the destination link succeeds, and rolls the link back if source removal fails. Unsupported filesystem semantics fail closed rather than reverting to replacing rename behavior. Targeted race regressions, complete Sandbox tests, and the full repository suite passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `08ad5b900e30a6755782a95ac28947c3f646fa15`.

- `go build ./...` passed

PR #29, opened 2026-08-22
fix(sandbox): surface incomplete glob traversal

### LINKED ISSUE #17: [sandbox] glob silently treats inaccessible subtrees as complete scans
## Summary

The `glob` tool swallows every `filepath.WalkDir` callback error. Directory errors return `filepath.SkipDir`; all other errors return `nil`. Neither branch records the skipped path or the error.

As a result, an unreadable directory (or even a root walk failure where `d == nil`) can be omitted while the tool reports `No files match pattern` with no warning and no `has_errors` metadata. This violates the SDK's fail-fast/observable-error contract and can make an agent conclude that a file does not exist when the scan was actually incomplete.

Relevant code:

- [`globWalkDir` callback discards `walkErr` without updating diagnostics](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_glob.go#L165-L177)
- [Only later per-file `Lstat`/open errors contribute to `skippedStat`; traversal errors never do](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_glob.go#L193-L233)
- [The final success/no-match response derives warnings solely from those later diagnostics](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_glob.go#L242-L303)
- [`grep` already demonstrates the expected behavior by recording walk failures](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_grep.go#L123-L151)

## Deterministic reproduction

The package already exposes `globWalkDir` as a test seam:

```go
func TestGlobReportsWalkFailure(t *testing.T) {
    root := t.TempDir()
    s, err := New(root)
    if err != nil {
        t.Fatal(err)
    }

    deps := tools.NewContainer()

    original := globWalkDir
    defer func() { globWalkDir = original }()
    globWalkDir = func(root string, fn fs.WalkDirFunc) error {
        // WalkDir uses d == nil for failures where it could not obtain an entry.
        return fn(filepath.Join(root, "unreadable"), nil, fs.ErrPermission)
    }

    ctx := context.Background()
    out, err := globTool().Execute(ctx, `{"pattern":"**/*.go"}`, deps)
    if err != nil {
        t.Fatalf("unexpected hard error: %v", err)
    }

    // Current result: "No files match pattern: **/*.go"
    // Current metadata: no has_errors, skipped_count, or warning_kind.
    if !strings.Contains(out.PlainText(), "WARN") {
        t.Fatalf("incomplete scan was reported as complete: %q", out.PlainText())
    }
}
```

A real filesystem reproduction is an unreadable subdirectory containing the only matching file. The callback receives an access error; `glob` skips the directory and reports no matches without disclosing that it could not scan it.

## Actual behavior

Traversal failures disappear from both the text result and tool metadata.

## Expected behavior

At minimum, every skipped traversal path must be counted and surfaced as `partial_scan_failure` with samples and an actionable warning. A failure on the requested root should preferably fail the tool outright.

## Impact

- False-negative file discovery.
- Agents may create duplicate files or take destructive actions based on an incorrect “not found” result.
- Permission and transient I/O problems are impossible for callers to distinguish from a genuinely empty result.
- `glob` behavior is inconsistent with `grep`, which records the same class of failures.

The callback explicitly receives a non-nil error and explicitly returns without storing it. The test seam makes the behavior deterministic; no platform-specific permission setup is required.

## Suggested fix

1. Record callback errors in a traversal diagnostic (separate permission vs other failures if useful).
2. Return a hard error when the requested root cannot be inspected.
3. For skipped descendants, set `has_errors`, `warning_kind=partial_scan_failure`, `skipped_count`, reason, and samples.
4. Add tests for root errors, unreadable directories, and mixed successful/failed traversal.

timwhitez:
## Summary

- record descendant `WalkDir` failures instead of silently treating the scan as complete
- return a hard error when the requested root itself cannot be inspected
- expose traversal counts/samples through warning metadata
- include an actionable warning in both no-match and partial-match output

## Validation

- [x] deterministic descendant traversal failure regression
- [x] root inspection failure regression
- [x] targeted Sandbox tests
- [x] final clean diff review completed after temporary automation was removed

Closes #17


timwhitez (commented):
Second-pass review completed after CI and cleanup. I rechecked root versus descendant WalkDir error handling, partial-scan metadata aggregation, and warning rendering for both empty and non-empty result sets. The final diff is limited to glob traversal handling and deterministic regressions; targeted Sandbox and full repository tests passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `186bb1d91b7bde491b960c88cca436c3a206ca8e`.

- `go build ./...` passed

PR #30, opened 2026-08-22
fix(openai): canonicalize strict schema required order

### LINKED ISSUE #25: [openai] strict-schema required arrays are nondeterministic
## Summary

The OpenAI strict-tool-schema conversion constructs each generated `required` array by ranging over a Go `map[string]any` of properties without sorting the names. Go map iteration order is unspecified, so equivalent logical tool definitions can serialize to different request bodies across allocations, processes, or calls.

## Evidence

- [`makeStrictSchema`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/chat.go) appends property names to `all` directly inside `for name, propAny := range props`.
- That unsorted slice is then assigned to `schema["required"]`.
- Recursive object conversion uses the same helper, so nested required arrays are also unstable.

Although array order is semantically irrelevant to JSON Schema validation, it is byte- and token-significant in the serialized model request.

## Deterministic reproduction

Repeatedly allocate an equivalent schema, convert it, marshal its `required` value, and collect the resulting byte strings:

```go
seen := map[string]struct{}{}
for i := 0; i < 1000; i++ {
    schema := freshSchemaWithProperties("alpha", "beta", "gamma", "delta")
    strict := makeStrictSchema(schema)
    body, _ := json.Marshal(strict["required"])
    seen[string(body)] = struct{}{}
}
if len(seen) != 1 {
    t.Fatalf("strict schema has %d required-array orders: %v", len(seen), seen)
}
```

Current behavior observes multiple orders.

## Actual behavior

Identical logical inputs can produce different serialized OpenAI request bodies.

## Expected behavior

Strict-schema conversion should produce one canonical representation at every object level.

## Impact

- Request hashes, signatures, recording/replay, and golden tests are unstable.
- Different token sequences reduce deterministic prompt-prefix/request-cache reuse.
- Diffs and diagnostics contain noise despite no semantic schema change.

## Suggested fix

Sort generated property-name lists before assigning `required`; recursive calls then canonicalize nested objects as well.

## Regression verification

Independently allocated top-level and nested schemas must serialize identically over many iterations.

timwhitez:
## Summary

- sort the generated strict-schema `required` array before request serialization
- apply the same canonical order recursively to nested object schemas
- add repeated deterministic checks for top-level and nested properties

## Validation

- [x] focused OpenAI strict-schema tests
- [x] final branch contains only the source change and regression test
- [x] second-pass diff review completed after temporary workflow removal

The validation workflow's final commit step raced with the already-created source commit; all focused and full test steps completed successfully before that administrative failure.

Closes #25


timwhitez (commented):
Second-pass review completed. I verified the canonical sort occurs before constructing `required`, recursive object conversion reuses the same helper, and the regression repeatedly checks top-level and nested order. Focused and full repository tests passed; the only workflow failure was a duplicate final commit attempt after the validated source commit already existed. The final diff is clean and contains no temporary automation. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `d6110d75282275e417530dc9a6875234ba63717c`.

- `go build ./...` passed

PR #32, opened 2026-08-22
fix(agent): reject duplicate exact tool names

### LINKED ISSUE #23: [agent] duplicate exact tool names expose conflicting definitions and execute the last registration
## Summary

`agent.New` accepts multiple tools with the same exact name. The ordered `tools` slice retains every definition for provider requests, while `toolMap` stores only one entry per name and silently overwrites earlier registrations.

The model can therefore receive multiple conflicting schemas/descriptions for one function name, but a returned call executes only the last registered handler.

## Evidence

- [`agent.New`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go) checks only that `Tool.Name` is non-empty, then assigns `toolMap[t.Name] = t`.
- The Agent separately retains `append([]tools.Tool(nil), cfg.Tools...)`, so duplicate definitions remain provider-visible.
- Exact-name resolution reads the overwritten map entry.
- Normalized-name collision behavior has explicit stable-order tests, but exact duplicates have no validation or documented override contract.

## Deterministic reproduction

Register:

- tool A: name `lookup`, schema `{query: string}`, handler A;
- tool B: name `lookup`, schema `{id: integer}`, handler B.

Construct an Agent and capture the provider request.

### Actual

- Construction succeeds.
- The provider receives two definitions named `lookup` with incompatible schemas.
- A model call named `lookup` resolves to handler B only.

### Expected

Construction should reject duplicate exact names before any provider invocation, or implement an explicit override API that replaces both the provider definition and execution entry atomically.

## Impact

- Providers may reject the request because function names are not unique.
- The model can follow schema A while the SDK invokes handler B.
- Registration order silently changes behavior.
- A safer/confirmation-gated implementation can be accidentally shadowed by a later duplicate.

## Suggested fix

During construction, track exact names and return an actionable error containing the duplicate name and registration positions. Keep normalized alias/collision handling as a separate policy.

## Regression verification

Constructing an Agent with two exact duplicate names must fail before any provider request or tool execution.

timwhitez:
## Summary

- reject duplicate exact tool names during `Agent` construction
- report both registration positions in the constructor error
- prevent the provider from receiving ambiguous duplicate definitions
- add a regression that verifies construction fails before any provider call

## Validation

- [x] focused duplicate-name constructor test
- [x] final branch contains no temporary validation workflow
- [x] second-pass diff review completed

The later `action_required` event is the workflow's bot-recursion guard; the preceding user-triggered validation run completed successfully.

Closes #23


timwhitez:
Second-pass review completed. I verified exact duplicate names are rejected before provider invocation, the error identifies both registration positions, and normalized-name collisions/alias behavior remain unchanged. Focused regression and the full repository suite passed in workflow run `32587849149`; the final diff contains only `agent.go` and its regression test. No merge or auto-merge performed.

timwhitez (commented):
Second-pass review completed. The constructor now rejects exact duplicate names before building either the provider-facing tool slice or the execution map, eliminating the previous schema/handler mismatch. The error identifies both registration positions, and the focused test verifies no provider invocation occurs. Focused and full repository tests passed; temporary automation is absent. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `36044304dc68b1b2ceb48e5c8b54480c1bfef74f`.

- `go build ./...` passed

PR #35, opened 2026-08-22
fix(agent): reject overlapping turns

### LINKED ISSUE #22: [agent] concurrent QueryStream calls mix turn history and cancellation state
## Summary

A single `Agent` accepts overlapping `QueryStream` / `QueryStreamWithSteering` calls without serializing or rejecting them. Each turn appends its input to shared history before taking the provider-request snapshot, and all turns share one active-stage cancellation slot.

## Evidence

- [`QueryStreamWithSteering`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go) starts a goroutine for every call and has no turn-level admission guard.
- Every turn appends its user input to the same `messages` slice, then later calls `Messages()` to build the provider request.
- The Agent has one shared `activeStageCancel` / generation slot for provider and tool stages.
- Per-output backpressure cancellation bookkeeping does not isolate shared history, compaction, steering, active-stage cancellation, or accounting sequence state.

Locking individual slice mutations prevents memory corruption, but it does not make the whole turn atomic or isolated.

## Deterministic reproduction outline

Use a provider stub with a two-party barrier:

1. Start `QueryStream(..., "alpha")`.
2. Start `QueryStream(..., "beta")` before either invocation proceeds.
3. Let both goroutines append their user message.
4. Release the provider barrier and capture each `InvokeRequest.Messages` value.

At least one—and commonly both—provider requests contains both newly submitted prompts. Assistant responses are then appended according to scheduler order rather than turn order.

A second test can make two provider stages active, interrupt one, and observe that the shared active-stage cancel function may refer to the other turn's stage.

## Actual behavior

Overlapping callers can see each other's in-flight prompts, mutate one shared turn history nondeterministically, and interfere with cancellation/steering.

## Expected behavior

The concurrency contract must be explicit and enforced. Either:

- turns on one Agent are serialized in submission order; or
- a second overlapping turn is rejected immediately with a typed busy error.

One caller's prompt, steering, cancellation, provider response, and accounting must never enter another caller's turn.

## Impact

- Cross-request context contamination and possible data disclosure.
- Nondeterministic history and replay behavior.
- Cancellation/steering cross-talk.
- Compaction and accounting may be attributed to the wrong logical turn.

## Suggested fix

Add turn-level admission control and scope active-stage cancellation to the admitted turn. A context-aware gate should let a queued caller cancel before it mutates history.

## Regression verification

Run two barrier-controlled queries concurrently and assert:

- each provider invocation contains only history committed before that turn plus its own new input;
- responses commit in the documented order;
- canceling one queued/active turn never cancels the other.

timwhitez:
## Summary

- enforce one active turn per `Agent` with synchronous admission before the turn goroutine starts
- return a structured terminal `agent_busy` error for overlapping submissions
- prevent rejected input from entering shared history or provider/tool execution
- release admission before closing the completed event stream
- document the concurrency contract and add deterministic/race regressions

## Validation

- [x] overlapping input/history isolation regression
- [x] 16-caller single-admission regression
- [x] post-completion admission-release regression
- [x] final source-only diff and lifecycle-order review

Validation run: `32590345561`.

Fixes #22


timwhitez:
Second-pass concurrency review completed after the final source-only cleanup. Admission uses a synchronous CAS before any history mutation/provider invocation; rejected submissions receive a buffered terminal `agent_busy` event without starting a goroutine; cleanup unregisters the turn and releases admission before the stream closes, so observing closure guarantees the next turn can enter. The regression verifies rejected input never enters history, only one provider call occurs across 16 simultaneous submissions, admission is restored after completion, and all focused tests (`-count=20`), race tests (`-count=10`), and the full repository suite passed in run `32590345561`. Final diff contains only source, tests, and documentation. No merge or auto-merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `347c67b59d97137324e2920d34dcace3e8dfe153`.

- `go build ./...` passed

PR #31, opened 2026-08-22
fix(agent): deep-clone all history API boundaries

timwhitez:
## Summary

- add canonical deep-clone helpers for messages, content blocks, pointer fields, tool calls, and thought-signature bytes
- take ownership of `InitialMessages` and `ReplaceHistory` input
- return an independent graph from `Messages`
- add mutation-isolation regressions for every mutable nested field

## Validation

- [x] constructor, getter, and replacement aliasing regressions
- [x] targeted Agent tests
- [x] focused race tests
- [x] full repository test suite
- [x] final patch review after temporary automation removal

Validation run: `32587087287`. The final branch contains only the three intended source/test files. A second-pass review verified all currently mutable nested fields are independently owned.

Closes #10

PR #36, opened 2026-08-22
fix(compaction): make checkpoint identities immutable snapshots

### LINKED ISSUE #12: [compaction] checkpoint identity can be invalidated through aliases to the source message graph
## Summary

`NewCompactionCheckpoint` computes a content-derived `checkpoint_id`, but it copies only the outer `[]llm.Message` slice. Nested message state (`Content.Blocks`, block pointers, `ToolCalls`, and `ThoughtSig` byte slices) remains shared with the caller's source messages.

Modifying the source message graph after construction therefore changes the checkpoint payload without changing its already-computed ID. The Agent commit path immediately hands that object to the configured writer without a final `Validate`, so a writer can persist a payload whose identifier no longer describes its contents if the source aliases are modified concurrently or retained by another component.

Relevant code:

- [`NewCompactionCheckpoint` shallow-copies only the outer message slice and then hashes it](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/compaction/runtime_checkpoint.go#L31-L48)
- [`Validate` correctly detects the resulting hash mismatch, but is not automatically applied by the constructor after aliases can change](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/compaction/runtime_checkpoint.go#L51-L79)
- [`llm.Message` contains nested slices, byte slices, and pointer-bearing content blocks](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/types.go#L35-L113)
- [The Agent constructs the checkpoint and passes it directly to `CheckpointWriter` without a final validation/deep snapshot](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L3840-L3875)

## Reproduction

```go
messages := []llm.Message{{
    Role: llm.RoleUser,
    Content: llm.Content{Blocks: []llm.ContentBlock{{
        Type: "text",
        Text: "original",
    }}},
}}
res := compaction.Result{Compacted: true}

checkpoint, err := compaction.NewCompactionCheckpoint(messages, res)
if err != nil {
    t.Fatal(err)
}
originalID := checkpoint.CheckpointID

// Mutate the constructor input, not the returned checkpoint.
messages[0].Content.Blocks[0].Text = "changed after hashing"

if checkpoint.Messages[0].Content.Blocks[0].Text != "changed after hashing" {
    t.Fatal("expected source alias to demonstrate the bug")
}
if checkpoint.CheckpointID != originalID {
    t.Fatal("stored ID unexpectedly changed")
}
if err := checkpoint.Validate(); err == nil || !strings.Contains(err.Error(), "hash mismatch") {
    t.Fatalf("Validate = %v, want hash mismatch", err)
}
```

The same mutation works through `ToolCalls[0].ThoughtSig`, `ContentBlock.ImageURL`, and `ContentBlock.Source`. A race-enabled variant can mutate the source while a recording writer serializes the checkpoint.

## Actual behavior

The returned checkpoint is not an isolated snapshot. Its payload can change through aliases that are outside the checkpoint object, while `checkpoint_id` and `Result.CheckpointID` stay fixed.

## Expected behavior

A constructor for a content-addressed, replayable checkpoint must take ownership of an immutable snapshot: subsequent mutations of the input graph must have no effect on the returned checkpoint.

## Impact

- A persisted checkpoint can fail its own integrity validation during recovery.
- The event/checkpoint ID may reference bytes different from those originally hashed.
- Concurrent source mutation can introduce a data race in checkpoint serialization.
- Recovery can fail closed after the Agent has already committed related ledger metadata, forcing rollback/full rebuild paths.

A Go struct/slice copy does not clone nested slice storage or pointers. The deterministic example mutates only the source argument and causes `checkpoint.Validate()` to fail. No model, filesystem timing, or platform behavior is involved.

## Suggested fix

Create and use a canonical deep-clone helper for `llm.Message` and every nested mutable field before computing the checkpoint ID. The same helper should be shared with the Agent history boundaries.

As defense in depth:

- call `checkpoint.Validate()` immediately before invoking the writer;
- document that writers must serialize synchronously or take their own deep copy before returning;
- add mutation-isolation tests for content blocks, image/document pointers, tool calls, and thought-signature bytes;
- add a race test with a recording writer to ensure the checkpoint graph is independent of the source.

timwhitez:
## Summary

- deep-clone the full message graph before calculating a checkpoint ID
- prevent source-history mutations from changing persisted checkpoint payloads
- verify content blocks, pointer fields, tool calls, and thought signatures remain isolated
- confirm the checkpoint still validates after source mutation

## Dependency

This PR is intentionally stacked on #31, which introduces the canonical `llm.CloneMessages` helper. Its diff against that branch contains only the checkpoint-specific change and regression test.

## Validation

- [x] deterministic source-alias mutation regression
- [x] targeted compaction tests
- [x] final clean stacked-diff review after temporary verification removal

Closes #12


timwhitez (commented):
Second-pass review completed after CI and cleanup. The checkpoint takes a canonical deep clone before identity hashing, so later mutations of block slices, pointees, tool calls, or signature bytes cannot change the stored payload or invalidate the original hash. The stacked diff is limited to the clone call and focused regression; targeted and full repository tests passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `be36a9f3445d6993495df7f5bdeff4b82e93013f`.

- `go build ./...` passed

PR #39, opened 2026-08-22
fix(agent): persist compaction checkpoints outside history lock

### LINKED ISSUE #13: [agent] asynchronous compaction invokes the host checkpoint writer while holding the history mutex
## Summary

When an asynchronously produced compaction result is applied, `Agent.applyPendingCompaction` acquires `a.mu` and keeps it held while calling `CommitCompactionCheckpoint`. That method synchronously invokes host-provided persistence callbacks, including `CheckpointWriter.SaveCompactionCheckpoint`.

A checkpoint writer that calls any Agent method needing the history mutex—most naturally `Messages()` to attach current history to its persistence/audit record—deadlocks permanently. Slow storage also blocks all history readers/writers for the full I/O duration.

Relevant code:

- [`applyPendingCompaction` acquires `a.mu`, constructs the merged history, and calls `CommitCompactionCheckpoint` before unlocking](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L3730-L3787)
- [`CommitCompactionCheckpoint` synchronously calls the configured checkpoint writer](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L3840-L3875)
- [`Messages` acquires the same non-reentrant `a.mu`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L454-L460)
- [`CompactionCheckpointWriter` is an arbitrary host callback](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/compaction/runtime_checkpoint.go#L21-L29)

## Reproduction

A deterministic package-level regression test can install a pending compacted result and use a writer that reads Agent history:

```go
func TestApplyPendingCompactionCheckpointWriterCanReadMessages(t *testing.T) {
    var a *Agent
    writerEntered := make(chan struct{})

    compCfg := &compaction.Config{
        Enabled: true,
        CheckpointWriter: compaction.CompactionCheckpointWriterFunc(
            func(ctx context.Context, cp compaction.CompactionCheckpoint) error {
                close(writerEntered)
                _ = a.Messages() // waits forever for a.mu held by applyPendingCompaction
                return nil
            },
        ),
    }

    a, _ = New(Config{LLM: fakeModel{}, Compaction: compCfg})
    a.messages = []llm.Message{llm.NewUserMessage("before")}
    a.pendingCompaction = &pendingCompaction{
        messages:    []llm.Message{llm.NewUserMessage("after")},
        snapshotLen: 1,
        result: compaction.Result{
            Compacted: true,
            // include whatever valid telemetry/checkpoint fields the test helper requires
        },
    }

    done := make(chan struct{})
    go func() {
        a.applyPendingCompaction(nil)
        close(done)
    }()

    <-writerEntered
    select {
    case <-done:
        // fixed
    case <-time.After(500 * time.Millisecond):
        t.Fatal("applyPendingCompaction deadlocked in checkpoint writer")
    }
}
```

The same class of deadlock applies if a persistence callback logs through code that indirectly calls `Messages`, or takes another application lock in the inverse order.

## Actual behavior

The Agent calls untrusted/host-controlled persistence code inside its core history critical section. Re-entering the history API deadlocks, and ordinary slow I/O serializes every concurrent turn/history operation behind storage latency.

## Expected behavior

No external callback or potentially blocking persistence operation should run while `a.mu` is held. The Agent should snapshot and validate the state under lock, release the lock, perform persistence, then reacquire the lock only to conditionally publish the result.

## Impact

- Permanent Agent/session deadlock from a reasonable checkpoint-writer implementation.
- Lock-order deadlocks with host audit/database locks.
- All queries, steering, `Messages`, and history mutations can be blocked by slow checkpoint storage.
- Cancellation cannot release the deadlock because `context.Background()` is passed to the commit path.

Go's `sync.Mutex` is not reentrant. The call chain is synchronous and the lock is visibly held across the callback. A writer calling `a.Messages()` always waits for the mutex that cannot be released until that same writer returns.

## Suggested fix

Use an optimistic publish pattern:

1. Under `a.mu`, capture the current history generation/version and build a deep, immutable merged snapshot.
2. Release `a.mu`.
3. Persist the ledger/checkpoint using the caller/turn context rather than `context.Background()` where possible.
4. Reacquire `a.mu` and publish only if the history generation/expected prefix still matches; otherwise retain/schedule the pending compaction for a new merge.
5. Keep rollback callbacks outside `a.mu` as well.

Add a regression test whose writer calls `Messages()`, plus a concurrency test proving a blocked writer does not block unrelated history reads.

timwhitez:
## Summary

- build an immutable compaction candidate under the history lock
- release the lock before warning callbacks, token estimation, ledger writes, and checkpoint persistence
- publish only when the live history still matches the captured source snapshot
- roll back the deferred ledger transaction and retry if history changes during persistence
- cover checkpoint writer, warning sink, and token-estimator re-entry plus slow-writer/read concurrency

## Dependency

This PR is intentionally stacked on #31, which provides the canonical `llm.CloneMessages` helper used for immutable source snapshots. Its diff against that branch is limited to the compaction publication transaction and focused regressions.

## Validation

- [x] checkpoint-writer re-entry tests
- [x] warning-callback and token-estimator re-entry tests
- [x] slow-writer/history-read concurrency test
- [x] final source-only diff and lock-boundary review

Validation run: `32591552624`.

Closes #13


timwhitez (commented):
Final second-pass review completed after the expanded callback tests. The history mutex now protects only capture and conditional publication; host checkpoint I/O, ledger persistence, Warningf, and custom TokenEstimator execution occur outside it. Publication uses a deep source snapshot comparison and rolls back the deferred ledger transaction on concurrent history change. Focused tests repeated 20 times, race tests repeated 10 times, and the full repository suite passed in run `32591552624`. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `8afe914b07391f0ffa00d399246db651839ae15e`.

- `go build ./...` passed

PR #40, opened 2026-08-22
fix(execrunner): terminate complete Windows process trees

### LINKED ISSUE #1: [Windows] command timeouts kill only the parent process and leave descendants running
## Summary

The Windows `execrunner` implementation creates a new process group, but both the graceful-termination and force-kill paths call `os.Process.Kill()` on only the direct child process. Descendants spawned by the shell are not terminated.

Relevant code:

- [`configureProcessGroup` only sets `CREATE_NEW_PROCESS_GROUP`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/process_group_windows.go#L10-L18)
- [Both termination functions call `proc.Kill()`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/process_group_windows.go#L20-L32)
- [`Runner.Run` relies on those functions when the context expires](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/runner.go#L136-L166)

## Reproduction

On Windows, run a shell command through `execrunner.Runner` that starts a long-lived child process and then waits, for example a PowerShell process that starts another PowerShell process running `Start-Sleep 300`. Configure a short command timeout.

After `Run` returns with a timeout, enumerate the child PID or observe it in Task Manager. The direct shell process is gone, but its spawned descendant remains alive.

A deterministic regression test can use a small helper executable that writes its PID to a file, spawns a second helper, and then blocks. After timeout, assert that neither PID is alive.

## Actual behavior

Only `cmd.Process` is terminated. `CREATE_NEW_PROCESS_GROUP` changes console/process-group behavior, but `os.Process.Kill()` is still scoped to that single process and does not terminate the descendants.

## Expected behavior

A timed-out command must terminate the complete process tree started for that invocation on every supported platform.

## Impact

- Timed-out tool calls can leave arbitrary background processes running.
- Descendants can retain inherited stdout/stderr handles and delay or corrupt command-output finalization.
- Repeated timeouts leak processes and system resources.
- Untrusted commands can intentionally escape the timeout boundary by spawning a child.

There is no Windows Job Object, descendant enumeration, or tree-scoped termination in the implementation. Both timeout stages invoke the exact same single-process `Kill()` call.

## Suggested fix

Create the command inside a Windows Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, assign the spawned process to it immediately after start, and terminate/close the job on cancellation. Add a Windows-specific process-tree regression test that verifies both the parent and a spawned grandchild are gone after timeout.

timwhitez:
## Summary

- start Windows commands suspended so no descendant can escape before containment
- attach the process to a Job Object configured with `KILL_ON_JOB_CLOSE`, then resume it
- terminate the complete Job Object on timeout/cancellation instead of killing only the direct shell
- fail closed if Job Object creation, assignment, or resume fails
- retain existing Unix process-group behavior through the same lifecycle abstraction
- verify a real Windows parent and descendant both disappear after timeout

## Validation

- [x] Windows parent/descendant process-tree regression
- [x] Windows Execrunner tests repeated 5 times
- [x] Ubuntu Execrunner tests repeated 5 times
- [x] full repository test suite on Ubuntu
- [x] patch hygiene and UTF-8 portability on both runners
- [x] final source-only cross-platform diff review

Validation run: `32636885083`.

Closes #1


timwhitez (commented):
Final cross-platform review completed after the Windows retry. Starting suspended closes the Start→Assign escape window; Job Object assignment happens before `NtResumeProcess`, `KILL_ON_JOB_CLOSE` covers surviving descendants on every exit path, and assignment/resume failures kill and reap the suspended process. Windows and Ubuntu Execrunner suites passed five repetitions, with the full repository suite passing on Ubuntu in run `32636885083`. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `aeda4e790113fa01ff07db87e54226b394c0f377`.

- `go build ./...` passed

Additional check: Windows/amd64 test-binary generation and full cross-build passed locally.

PR #41, opened 2026-08-22
fix(execrunner): abort canonical artifacts when streams remain open

### LINKED ISSUE #2: Canonical stdout/stderr artifacts are committed as complete while escaped descendants can still write
## Summary

When a timed-out command has a descendant that escaped the process group and still holds stdout/stderr open, `Runner.Run` stops waiting after `KillWaitGrace` but immediately commits the canonical stream artifacts. The background `cmd.Wait()` goroutine and stream writers remain active, so bytes can arrive after the artifact has been marked complete and recoverable. Those late bytes are silently discarded by `canonicalStreamCapture.write` because `finished` is already true.

Relevant code:

- [`Runner.Run` deliberately returns while `cmd.Wait` can still be blocked](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/runner.go#L151-L188)
- [It then calls `canonical.finish` unconditionally and snapshots the manifests](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/runner.go#L190-L211)
- [`write` discards every byte received after `finished` is set](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/canonical_stream.go#L157-L163)
- [`finish` commits the object and accepts only a manifest claiming `complete` and `recoverable`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/canonical_stream.go#L185-L207)

## Reproduction

On Unix, run a shell command that:

1. writes an early stdout prefix;
2. starts a `setsid` child that inherits stdout, sleeps longer than `KillWaitGrace`, then writes a suffix;
3. keeps the original command alive long enough to hit the runner timeout.

For example, a regression test can invoke a shell equivalent to:

```sh
printf early
setsid sh -c 'sleep 0.2; printf late' &
sleep 30
```

Use a short command timeout and a shorter `KillWaitGrace`, with a recording `ArtifactStreamSink`.

`Run` returns with `KillProcessTimedOut=true` and a canonical stdout manifest containing only `early`. After return, the escaped process writes `late`, but the already-finished canonical capture drops it. The manifest still reports the prefix as a complete and recoverable stdout object.

## Actual behavior

A prefix of a stream that is still open is committed as the complete canonical stream. There is no diagnostic saying the object is partial or abandoned.

## Expected behavior

Canonical artifacts must not be marked complete/recoverable until the corresponding output stream has reached EOF. If the runner intentionally gives up waiting, the open stream should be aborted or represented explicitly as incomplete.

## Impact

- Audit/recovery consumers receive a false completeness guarantee.
- Output that arrives after the timeout boundary is silently lost from the canonical object.
- Hash and byte-count metadata describe only a prefix while callers are told it is the complete process stream.
- The behavior is especially dangerous for forensic, accounting, and resumability uses of canonical artifacts.

The code explicitly preserves the stream writer for later background writes when `cmd.Wait` has not returned, but commits the canonical stream before those writes finish. The capture's `finished` guard deterministically drops every subsequent write.

## Suggested fix

When `killWaitTimedOut` is true, do not call `Commit` on either canonical stream. Abort it and emit an actionable incomplete-stream diagnostic, or move finalization into the background owner that closes the collector after `cmd.Wait` actually returns. If partial artifacts are useful, extend the manifest contract to represent `complete=false`/`recoverable=false` rather than validating the prefix as complete. Add the escaped-`setsid` regression described above.

timwhitez:
## Summary

- do not commit stdout/stderr canonical objects when `cmd.Wait` exceeded `KillWaitGrace`
- abort open stream writers and emit an explicit incomplete/non-recoverable diagnostic
- preserve the bounded preview and process timeout result without claiming a complete artifact
- verify an escaped `setsid` descendant cannot cause a prefix to be committed as a complete object, even after late output arrives

## Validation

- [x] escaped-descendant/late-write regression
- [x] complete execrunner package tests

Closes #2


timwhitez (commented):
Second-pass review completed after CI and cleanup. `write`, `finish`, and `abortIncomplete` share the same lock; once an incomplete stream is marked finished, late writes cannot create or extend an artifact. Open writers are aborted, no manifest is exposed, and the escaped-descendant test verifies no delayed commit occurs after `Run` returns. Execrunner and full repository tests passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `d547d623ba4187b283d7964b06584c06f5a6ea3f`.

- `go build ./...` passed

PR #42, opened 2026-08-22
fix(sandbox): pin webfetch sockets to validated DNS results

### LINKED ISSUE #3: [security] webfetch destination filtering is bypassable via DNS rebinding
## Summary

`webfetch` resolves a hostname during destination validation and rejects private/special-purpose addresses, but it discards the validated addresses. The subsequent `http.Client.Do` performs a new DNS lookup through the default transport. A hostname that returns a public address for validation and a private/metadata address for the connection can therefore bypass the SSRF guard.

Relevant code:

- [The request target is validated before the request](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L52-L65)
- [Validation resolves and classifies all returned addresses, then returns without retaining them](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L158-L203)
- [The client uses the default transport, which resolves the hostname again during dialing](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L94-L122)
- [Redirects repeat the same validate-then-resolve-again pattern](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L97-L108)

## Reproduction

Use an attacker-controlled DNS name with a very short TTL or an authoritative server that rotates answers:

1. The first lookup returns a permitted public address.
2. The next lookup returns `169.254.169.254`, `127.0.0.1`, an RFC1918 address, or another denied destination.
3. Invoke `webfetch` for that hostname and approve the confirmation.

`validateWebfetchHostDestination` accepts the first answer. `http.Transport` then resolves the same hostname again and connects to the second answer.

A deterministic regression test can use a local DNS server whose first response is public and second response is loopback, together with a loopback HTTP listener. The request must be rejected and the listener must receive no connection.

## Actual behavior

The policy validates one DNS result but the socket can be opened to a different result.

## Expected behavior

The address that is dialed must be one of the addresses that passed classification, and the connected peer must remain inside the allowed set for the entire redirect chain.

## Impact

This permits SSRF to:

- cloud instance-metadata services;
- localhost services;
- private network services;
- link-local and other explicitly denied destinations.

The existing comprehensive address classifier does not protect the actual connection when DNS can change between validation and dialing.

The client has no custom `Transport`/`DialContext`, no validated-address pinning, and no post-connect peer-address check. The validation lookup and the dial lookup are separate operations by construction.

## Suggested fix

Resolve once after confirmation, classify the result, and pass the accepted IP set into a custom `http.Transport.DialContext`. Dial an accepted IP directly while preserving the original hostname for the HTTP `Host` header and TLS SNI/certificate verification. Apply the same process independently to every redirect target. Optionally verify `RemoteAddr` after connect as defense in depth. Add a rotating-DNS regression test.

## Closed-loop audit follow-up (2026-08-24)

A second transport review found an additional path to the same destination-to-socket violation: cloning a process-global `*http.Transport` can inherit `DialTLSContext`, legacy `DialTLS`, proxy routing, and caller TLS policy. For HTTPS, a TLS dial hook can bypass the validated `DialContext` entirely even after DNS pinning is added.

The fixing PR therefore must use a package-owned transport with no inherited proxy/TLS dial hooks or TLS configuration, and must prove hostile process-global transport state is not reused. This is the same exact-socket invariant described by this Issue, so no duplicate Issue was opened.

timwhitez:
## Summary

- classify the exact DNS result used for each socket connection
- dial an approved literal IP directly while preserving the original HTTP Host and TLS SNI
- disable proxy and process-global custom RoundTripper paths that could resolve independently
- re-apply socket-bound validation to redirect destinations
- use a fixed package-owned transport so inherited `DialTLSContext`, `DialTLS`, proxy, and TLS-policy hooks cannot bypass the validated `DialContext`
- keep test injection behind a request-execution seam that cannot alter production transport policy

## Closed-loop review

A second-pass security review found that cloning a process-global `*http.Transport` could retain TLS dial hooks that bypass `DialContext` for HTTPS. Commit `3c5515c` replaces that clone with a package-owned transport and adds regression coverage for inherited proxy, TLS-dial, and TLS-policy hooks.

## Local validation

Audited locally at exact head `3c5515ca1da0dece83b24d42a13c6045f0d18c67` with Go 1.24.5 on 2026-08-24; no GitHub Actions were invoked:

- [x] `gofmt` clean
- [x] `go vet ./...`
- [x] focused Sandbox tests
- [x] focused Sandbox race tests
- [x] `go build ./...`
- [x] final source-only security diff review

Closes #3


timwhitez (commented):
Final security review completed after the fixture and transport-policy corrections. The production client owns a controlled `*http.Transport`, disables proxy resolution, classifies the exact DNS result immediately before dialing, and passes only a validated literal IP to the socket dialer while preserving URL host/SNI. A process-global custom RoundTripper cannot bypass the policy. Redirect targets use the same validator. Focused tests repeated 20 times, race tests repeated 10 times, complete Sandbox tests, and the full suite passed in run `32638343812`. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `3c5515ca1da0dece83b24d42a13c6045f0d18c67`.

- `go build ./...` passed

PR #43, opened 2026-08-22
fix(sandbox): reject overflowing and unreasonable timeouts

### LINKED ISSUE #5: [sandbox] oversized timeout values overflow time.Duration and silently disable limits
## Summary

The `bash` and `webfetch` tools accept timeout seconds as an unbounded `int` and convert them with `time.Duration(timeout) * time.Second`. On 64-bit builds, values above `9223372036` seconds overflow `time.Duration` to a non-positive value. Both downstream implementations interpret a non-positive duration as “no timeout”, so a model-supplied oversized positive timeout disables the execution/network bound instead of being rejected.

Relevant code:

- [`bashArgs.Timeout` is an `int` and only non-positive input is defaulted](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_bash.go#L18-L21)
- [`bash` multiplies it without an overflow/range check](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_bash.go#L58-L87)
- [`execrunner.Run` installs a deadline only when the resulting duration is positive](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/execrunner/runner.go#L83-L105)
- [`webfetch` performs the same unchecked conversion](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L66-L103)

## Reproduction

On a 64-bit build, supply:

```json
{"timeout": 9223372037}
```

For `bash`, use a command that never exits. For `webfetch`, target a server that accepts the connection and never produces a response.

`time.Duration(9223372037) * time.Second` wraps to a negative duration. `execrunner.Run` skips `context.WithTimeout`, and `http.Client` treats `Timeout <= 0` as no timeout. The operation remains bounded only by an unrelated parent context, if one exists.

A unit test can calculate the converted duration and exercise both tools with an independently cancellable parent context; assert that the oversized value is rejected before confirmation/execution rather than treated as unlimited.

## Actual behavior

A very large positive timeout disables the intended tool-level timeout.

## Expected behavior

Timeout input must be validated before conversion. Out-of-range or unreasonable values should return an actionable argument error; they must never wrap into an unlimited duration.

## Impact

- Commands or network requests can run indefinitely.
- An untrusted model/tool caller can bypass the default 30-second safety bound.
- Stuck calls retain processes, sockets, goroutines, and confirmation state.

The overflow threshold follows directly from `math.MaxInt64 / int64(time.Second)`. The resulting non-positive value reaches code paths that explicitly mean “do not configure a timeout”. No later validation restores the bound.

## Suggested fix

Use a checked conversion such as:

```go
const maxTimeoutSeconds = int64(math.MaxInt64 / int64(time.Second))
if timeout <= 0 || int64(timeout) > maxTimeoutSeconds || timeout > configuredReasonableMaximum {
    return an actionable validation error
}
d := time.Duration(timeout) * time.Second
```

Prefer a shared helper for all sandbox tools and apply a practical upper bound (for example minutes/hours rather than centuries). Add boundary tests at the maximum accepted value and the first rejected value.

timwhitez:
## Summary

- centralize checked seconds-to-`time.Duration` conversion for sandbox tools
- reject values outside `time.Duration` range and values above a 24-hour practical maximum
- validate Bash and WebFetch timeout arguments before confirmation or execution
- preserve the existing default behavior for omitted/non-positive timeout fields
- add maximum, maximum+1, overflow, and pre-confirmation regressions
- keep the WebFetch regression independent of external DNS by using the package's deterministic resolver with the reserved `example.test` domain

## Local validation

Audited locally with Go 1.24.5 on 2026-08-24; no GitHub Actions were invoked:

- [x] `gofmt` clean
- [x] `go vet ./...`
- [x] `go test -count=1 ./sdk/tools/sandbox`
- [x] `go test -race -count=1 ./sdk/tools/sandbox`
- [x] `go build ./...`
- [x] final source-only diff review

The first local run exposed that the WebFetch regression used `example.com` and therefore failed in an offline environment before reaching its timeout assertion. Commit `218160c` switches the test to the existing deterministic resolver and `example.test`; the complete gate then passed.

Closes #5


timwhitez (commented):
Final review completed after successful validation. The checked conversion precedes confirmation and execution, distinguishes representational overflow from the practical 24-hour policy ceiling, and preserves existing default handling for non-positive fields. Targeted boundary tests, complete Sandbox tests, and the full repository suite passed in run `32590861705`. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `218160cbb0411ea728fb708511bfbc52b5b44d96`.

- `go build ./...` passed

Closed-loop correction: the original regression depended on external DNS. The test now uses the deterministic in-process resolver with `example.test`; focused/race/full/build gates all passed after the update.

PR #44, opened 2026-08-22
fix(openai): release abandoned streaming producers on cancellation

### LINKED ISSUE #6: [openai] streaming goroutines leak when callers stop consuming the event channel
## Summary

Both OpenAI streaming adapters publish events with unconditional sends to a bounded channel. If a caller stops reading after the 128-slot buffer fills, the producer goroutine blocks forever on `out <- event`. Cancelling the request context does not unblock a channel send, so the goroutine cannot close the response body or the output channel.

Relevant code:

- [Chat streaming creates a 128-element channel and launches the producer](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/chat.go#L204-L216)
- [Chat emits all event types with direct channel sends, including inside the SSE callback](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/chat.go#L360-L435)
- [Responses streaming uses the same bounded-channel/direct-send pattern](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/responses.go#L389-L420)
- [Responses helper closures also send directly and cannot observe cancellation while blocked](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/responses.go#L535-L590)

## Reproduction

1. Start an `httptest.Server` that returns a valid SSE stream containing more than 128 text-delta events and keeps the connection open.
2. Call either OpenAI client's `InvokeStream`.
3. Do not read from the returned channel, or read one event and then stop.
4. Cancel the context after the server has produced enough events.
5. Assert that the producer exits, the response body is closed, and the event channel closes.

The assertions currently fail: the producer remains blocked on a send after the buffer fills. A goroutine profile shows it parked in `chan send` from `chat.go` or `responses.go`.

## Actual behavior

Cancellation only reaches network reads/backoff code. It cannot interrupt an already-blocked `out <- event`, so producer resources remain retained indefinitely.

## Expected behavior

Stopping consumption and cancelling the context must terminate the producer promptly and close the HTTP response body/channel.

## Impact

- Leaked goroutines and HTTP connections.
- Connection-pool exhaustion under repeated abandoned streams.
- Requests and provider response bodies remain retained in memory.
- Agent/session shutdown can hang or accumulate background work.

A send to a full Go channel is not context-aware. Every producer path uses direct sends, and there is no separate draining goroutine or close protocol that can release the blocked producer.

## Suggested fix

Route every event through a helper such as:

```go
func sendEvent(ctx context.Context, out chan<- llm.StreamEvent, ev llm.StreamEvent) bool {
    select {
    case out <- ev:
        return true
    case <-ctx.Done():
        return false
    }
}
```

Return immediately when it reports cancellation. Apply it to success, error, usage, response-ID, text/thinking, tool-call, and completion events in both OpenAI adapters. Ensure response bodies are closed with a defer owned by the producer. Add regression tests for an unread full channel and a partially consumed channel.

timwhitez:
## Summary

- separate provider production from caller delivery with a shared forwarding layer
- preserve normal backpressure while the request is active
- switch to drain-and-drop mode after cancellation so blocked provider producers can exit
- close the public channel only after the internal producer has released the HTTP body
- cover both Chat Completions and Responses with unread-consumer regressions

## Validation

- [x] unread/full-channel cancellation regression for both OpenAI adapters
- [x] focused race test
- [x] complete OpenAI provider tests

Closes #6


timwhitez (commented):
Second-pass review completed after CI and cleanup. The public channel has a single owner; on cancellation the forwarder switches to drain-and-drop mode rather than returning, so internal producers cannot remain blocked and no producer can send to a closed public channel. Both adapters share the same implementation. Focused, race, provider-package, and full repository tests passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `4a977ef01d5ea5495a43dc15790d1e271e769506`.

- `go build ./...` passed

PR #45, opened 2026-08-22
fix(agent): require explicit terminal events from provider streams

### LINKED ISSUE #8: [agent] provider streams that close without StreamDoneEvent are accepted as successful completions
## Summary

The generic Agent streaming consumer treats closure of a provider event channel as successful completion regardless of whether the provider ever emitted `llm.StreamDoneEvent`. This makes the Agent-level contract vulnerable even after individual provider adapters are fixed: any custom or future `StreamingChatModel` can disconnect mid-response and have its partial output accepted as a completed turn.

Relevant code:

- [`StreamDoneEvent` only copies the stop reason; no completion flag is recorded](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L1889-L1892)
- [When the event channel closes, `invokeCompletionWithSteering` returns the accumulated completion with a nil error](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L1930-L1937)

## Reproduction

Implement a minimal fake `llm.StreamingChatModel` whose `InvokeStream` method:

1. returns a channel;
2. sends `llm.StreamTextDeltaEvent{Delta: "partial"}`;
3. closes the channel without sending `StreamDoneEvent` or `StreamErrorEvent`.

Run an Agent turn using that model. The Agent returns a completion containing `partial` with a nil error. At the normal loop level, that partial assistant output can then be appended to conversation history as a successful model turn.

The same reproduction works with an incomplete `StreamToolCallDeltaEvent`, making malformed tool-call arguments appear to be the final provider result.

## Actual behavior

Channel closure itself is treated as the successful end-of-stream signal.

## Expected behavior

A provider stream should complete successfully only after an explicit `StreamDoneEvent`. Channel closure before Done must return a typed incomplete-stream error while retaining the partial completion for diagnostics/history policy.

## Impact

- Custom provider disconnects and implementation bugs silently truncate responses.
- Partial tool-call names/arguments can be finalized and executed or persisted.
- Agent retry/recovery logic is bypassed because the close path returns nil.
- The generic abstraction provides no enforceable terminal-event contract to provider implementations.

The close branch unconditionally passes `nil` to `finishProviderStage`, and no state records whether `StreamDoneEvent` was observed. This is deterministic with a three-event fake provider and does not depend on transport behavior.

## Suggested fix

Track `sawDone` in `invokeCompletionWithSteering`:

- set it when a valid `StreamDoneEvent` is processed;
- on channel close, return success only when `sawDone` is true;
- otherwise return a typed `IncompleteStreamError` plus `partialCompletion()`;
- consider rejecting events received after Done as a provider-contract violation.

Add Agent-level contract tests for: normal Done then close, text then close without Done, metadata-only then close, error then close, and incomplete tool-call deltas then close.

timwhitez:
## Summary

- introduce a typed `llm.IncompleteStreamError`
- record whether the generic provider stream emitted `StreamDoneEvent`
- return the typed error when the event channel closes first while preserving partial text/tool-call data
- preserve valid metadata-only streams and existing provider errors
- add Agent-level contract tests independent of any specific provider adapter

## Validation

- [x] partial text without Done regression
- [x] incomplete tool-call without Done regression
- [x] Done, metadata-only, and provider-error contract tests
- [x] complete Agent package tests

Closes #8


timwhitez (commented):
Second-pass review completed after CI and cleanup. I verified terminal metadata buffered before visible content still sets the Done flag when flushed, metadata-only streams with Done remain valid, partial text/tool-call state is preserved on incomplete closure, and existing provider errors retain their type. Agent-package and full repository tests passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `4a22acdca50d03d19e33fd095d7f242ab808e359`.

- `go build ./...` passed

timwhitez (commented):
Independent audit result: PASS. I traced the generic stream state machine, including metadata buffering, explicit Done handling, partial text/tool-call preservation, provider-error precedence, and metadata-only completion. The typed IncompleteStreamError closes the reported contract gap without changing valid Done-terminated streams. No unresolved review threads or required checks remain. GitHub does not permit the PR author account to self-approve, so this is recorded as a completed independent audit comment.

PR #46, opened 2026-08-22
fix(retry): fail fast on static authentication failures

### LINKED ISSUE #9: [retry] 401/403 authentication failures are retried at both provider and agent layers
## Summary

The default OpenAI retry policy classifies HTTP 401 and 403 as retryable. After the provider exhausts its default 10 attempts, the Agent framework classifies the same `ProviderError` statuses as transient and repeats the entire provider invocation with its default 2-attempt policy. A static invalid API key or denied model permission can therefore generate up to 20 identical requests and several minutes of backoff before the real error reaches the caller.

Relevant code:

- [The OpenAI default retry policy uses 10 total attempts](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/retry_policy.go#L9-L31)
- [The OpenAI status classifier includes both 401 and 403](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/chat.go#L670-L686)
- [The Agent defaults to two framework-level attempts](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L274-L323)
- [The Agent retry loop invokes the complete provider call again](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L1991-L2054)
- [`retryableProviderStatus` independently includes 401 and 403](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/agent/agent.go#L4303-L4311)

## Reproduction

1. Start an `httptest.Server` that always returns HTTP 401 (repeat with 403) and increments a request counter.
2. Configure a default `openai.ChatClient` or `openai.ResponsesClient` against the server.
3. Construct an Agent with default retry settings.
4. Invoke one turn and record the number of HTTP requests.

With delays shortened through test configuration, the counter reaches 20: 10 provider attempts for the first framework attempt, then 10 more for the second.

The same direct provider call without the Agent still sends 10 requests for an unchanged authorization failure.

## Actual behavior

Permanent authentication/authorization failures are retried as transient at two independent layers.

## Expected behavior

401 and 403 should fail immediately unless a specific integration has an explicit credential-refresh/permission-propagation mechanism and opts into a narrowly scoped retry after that state changes.

## Impact

- Up to 20 identical unauthorized requests per Agent invocation.
- Default provider backoff can delay the error by several minutes.
- Authentication incidents create avoidable traffic and noisy security/audit logs.
- Interactive clients appear hung instead of reporting the actionable credential or permission error.

The request's `Authorization` value comes from the same static client field on every provider attempt. Neither the OpenAI retry loop nor the Agent retry loop refreshes credentials or changes authorization state. The framework itself labels its retry facility as being for transient failures, while separate error classification already identifies 401 as `auth` and 403 as `permission`.

## Suggested fix

- Remove 401 and 403 from `defaultRetryableStatus` and `retryableProviderStatus`.
- Preserve opt-in override support through `RetryableStatusCodes` for unusual gateways.
- If credential refresh is added later, retry only after a refresh callback reports that credentials actually changed, and avoid applying a second framework retry to an already-exhausted provider error.
- Add request-count regression tests for 401/403 at both direct-provider and Agent integration levels.

timwhitez:
## Summary

- remove HTTP 401/403 from the OpenAI default retry classifier
- remove the same permanent statuses from Agent framework retries
- preserve explicit provider opt-in through `RetryableStatusCodes`
- update legacy textual retry classification tests to treat auth failures as permanent
- verify direct Chat/Responses calls and full Agent turns issue exactly one request by default

## Validation

- [x] direct Chat/Responses 401/403 request-count tests repeated 20 times
- [x] Agent integration request-count tests repeated 20 times
- [x] explicit 401 provider retry override regression
- [x] textual rate-limit/server/non-retryable classifier tests repeated 20 times
- [x] full repository test suite passed twice
- [x] final source-only diff review

Validation run: `32637569675`.

Closes #9


timwhitez (commented):
Final review completed after synchronizing legacy retry expectations. Both provider-level and Agent-level default classifiers now treat 401/403 as permanent, while the explicit `RetryableStatusCodes` override remains effective inside the provider client. Real request-count tests cover Chat, Responses, and a full Agent turn; transient 408/409/425/429/5xx behavior remains covered. All focused tests repeated 20 times and two complete repository runs passed in `32637569675`. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `331f95ba98cfed11b5da4fa6eba04a502fdac8f1`.

- `go build ./...` passed

PR #48, opened 2026-08-22
fix(sandbox): decode UTF-16 and UTF-32 BOM text before reading

### LINKED ISSUE #19: [sandbox] read accepts UTF-16/UTF-32 BOM files but returns raw NUL-filled bytes
## Summary

The `read` tool treats UTF-16 and UTF-32 BOMs as proof that a file is textual, but it never decodes those encodings. After binary detection, the same file is rewound and passed directly to `bufio.Scanner`, which splits on the single byte `\n` and returns raw byte sequences as Go strings.

Consequently UTF-16/UTF-32 files are reported as successfully read while the content contains BOM bytes, embedded NULs, broken line boundaries, and often invalid UTF-8. The current regression test only checks that UTF-16 does not return an error; it never checks the returned text.

Relevant code:

- [`isBinaryData` returns false for any recognized UTF-8/16/32 BOM](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_read.go#L161-L186)
- [`read` then seeks to byte zero and uses the default byte-oriented `bufio.Scanner` without decoding](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_read.go#L75-L105)
- [`hasUnicodeTextBOM` explicitly recognizes UTF-16 LE/BE and UTF-32 LE/BE](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_read.go#L191-L215)
- [`TestReadToolAllowsUTF16WithBOM` asserts only `err == nil`, not decoded output](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_test.go#L2610-L2630)

## Deterministic reproduction

```go
func TestReadToolDecodesUTF16LE(t *testing.T) {
    root := t.TempDir()
    s, err := New(root)
    if err != nil {
        t.Fatal(err)
    }

    // BOM + "hi\nworld\n" encoded as UTF-16LE.
    raw := []byte{
        0xff, 0xfe,
        'h', 0, 'i', 0, '\n', 0,
        'w', 0, 'o', 0, 'r', 0, 'l', 0, 'd', 0, '\n', 0,
    }
    if err := os.WriteFile(filepath.Join(root, "utf16.txt"), raw, 0o644); err != nil {
        t.Fatal(err)
    }

    deps := tools.NewContainer()

    out, err := readTool().Execute(context.Background(), `{"file_path":"utf16.txt"}`, deps)
    if err != nil {
        t.Fatal(err)
    }
    got := out.PlainText()
    if strings.ContainsRune(got, '\x00') || !strings.Contains(got, "hi") || !strings.Contains(got, "world") {
        t.Fatalf("UTF-16 was not decoded: %q", got)
    }
}
```

Current output contains the BOM and embedded `\x00` bytes. The scanner also sees each UTF-16 newline as a lone byte delimiter, leaving the other half of the code unit attached to adjacent lines.

UTF-16BE and both UTF-32 variants fail similarly; UTF-32 introduces three NUL bytes around most ASCII characters.

## Actual behavior

The tool returns success with corrupted/garbled text.

## Expected behavior

Either:

1. Detect the BOM, decode the stream to UTF-8, strip the BOM, and then perform line/offset/output-limit processing; or
2. Reject unsupported encodings with an actionable error instead of claiming a successful read.

## Impact

- Agents receive incorrect source/config/document contents and may make invalid edits.
- Line numbers and offsets are unreliable for UTF-16/32 files.
- Tool output can contain invalid UTF-8 or large numbers of escaped NULs when serialized.
- The behavior contradicts the explicit decision to classify these encodings as text.

The reproduction uses a standards-compliant BOM and ASCII characters encoded in UTF-16LE. `bufio.Scanner` does no charset conversion; raw NULs are therefore guaranteed in the returned string. The existing test's missing output assertion explains why the implementation currently passes.

## Suggested fix

- Detect the exact BOM before scanning.
- Decode UTF-16 LE/BE and UTF-32 LE/BE to UTF-8 with strict malformed-sequence handling.
- Strip the BOM before line numbering.
- Apply byte/line limits to both encoded input and decoded output to preserve DoS protections.
- Extend tests to assert exact decoded content, line numbering, offsets, malformed sequences, and all four endian variants.

timwhitez:
## Summary

- detect UTF-8/16/32 BOM variants before line scanning
- strictly decode UTF-16LE/BE and UTF-32LE/BE to UTF-8
- reject truncated code units, unpaired surrogates, invalid UTF-8, and invalid Unicode scalar values
- strip the BOM before line numbering and preserve offset/limit behavior
- bound decoded output independently of the encoded input size

## Validation

- [x] exact-output tests for all four UTF-16/32 endian variants
- [x] offset/limit regression
- [x] malformed/truncated sequence regressions
- [x] complete Sandbox tests

Closes #19


timwhitez (commented):
Second-pass review completed after CI and cleanup. BOM detection checks UTF-32 before UTF-16, malformed code units/surrogates/scalars fail closed, decoded output remains bounded for worst-case UTF-16 expansion, and line numbering/offset/limit operate on decoded UTF-8. Targeted Unicode, complete Sandbox, and full repository tests passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `162890faaf0972827a40cd7eb10364645ce332b6`.

- `go build ./...` passed

PR #49, opened 2026-08-22
fix(sandbox): bound model-controlled read and grep counts

### LINKED ISSUE #20: [sandbox] unbounded read/grep numeric arguments can panic or exhaust the host
## Summary

Several model-controlled integer arguments are accepted without upper bounds and are used directly as slice capacities:

- `read.limit` becomes `make([]string, 0, limit)`.
- `grep.before` / `grep.context` become `make([]prevLine, 0, before)` for each scanned file.

A valid JSON integer near `math.MaxInt` therefore causes a deterministic `runtime error: makeslice: cap out of range`; smaller but still large values can force excessive allocation and terminate the process through OOM. The generic tool execution wrappers perform strict JSON decoding but neither validate these ranges nor recover panics.

Relevant code:

- [`readArgs.Limit` has no validation tag or maximum](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_read.go#L18-L24)
- [`read` uses a positive caller value directly as the output slice capacity](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_read.go#L84-L99)
- [`grepArgs` exposes unbounded `Before`, `After`, `Context`, and `MaxResults`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_grep.go#L19-L35)
- [`grep` clamps only negative values, not oversized positive values](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_grep.go#L65-L88)
- [Every scanned file allocates a previous-line buffer with capacity `before`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_grep.go#L181-L191)
- [`Tool.Execute` and `toolWithArgs` decode and invoke handlers without panic recovery or semantic numeric bounds](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/tool.go#L39-L88) and [here](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_helpers.go#L244-L266).

### `read.limit`

```go
func TestReadRejectsOversizedLimit(t *testing.T) {
    root := t.TempDir()
    if err := os.WriteFile(filepath.Join(root, "x.txt"), []byte("x\n"), 0o644); err != nil {
        t.Fatal(err)
    }
    s, _ := New(root)
    deps := tools.NewContainer()

    // On a 64-bit build this reaches make([]string, 0, math.MaxInt).
    _, _ = readTool().Execute(
        context.Background(),
        fmt.Sprintf(`{"file_path":"x.txt","limit":%d}`, int64(math.MaxInt)),
        deps,
    )
}
```

Current result on a normal 64-bit runtime is a panic similar to:

```text
runtime error: makeslice: cap out of range
```

### `grep.context`

```go
func TestGrepRejectsOversizedContext(t *testing.T) {
    root := t.TempDir()
    _ = os.WriteFile(filepath.Join(root, "x.txt"), []byte("needle\n"), 0o644)
    s, _ := New(root)
    deps := tools.NewContainer()

    _, _ = grepTool().Execute(
        context.Background(),
        fmt.Sprintf(`{"pattern":"needle","context":%d}`, int64(math.MaxInt)),
        deps,
    )
}
```

The first regular file reaches `make([]prevLine, 0, before)` and panics before scanning its contents.

Using values that are below the runtime's immediate `maxAlloc` check can instead request gigabytes of memory, making this an OOM vector rather than only a recoverable panic.

## Actual behavior

Syntactically valid tool arguments can panic inside `Tool.Execute` or consume unbounded memory.

## Expected behavior

All model-controlled counts, offsets, context windows, result limits, and timeouts should have explicit safe upper bounds. Oversized values should return an actionable validation error without allocating proportional memory.

## Impact

- A malformed or adversarial model tool call can terminate the agent host.
- Direct SDK consumers calling `Tool.Execute` are exposed even if a higher-level integration happens to recover panics.
- `grep.context` multiplies the allocation risk across files.
- An OOM termination is not reliably recoverable and can lose in-memory agent state.

The vulnerable allocations occur before the requested capacity is reduced by file size or output limits. Go's `make` behavior for impossible capacities is deterministic, and no upper-bound check exists in either handler or the shown execution wrappers.

## Suggested fix

1. Define conservative hard caps, for example `read.limit <= 10_000`, `grep context <= 1_000`, and bounded `max_results`.
2. Reject values above the cap rather than silently clamping, so callers know the request was changed.
3. Avoid preallocating directly from untrusted values; cap initial capacity independently of the logical maximum.
4. Add a last-resort panic boundary around tool handlers that converts ordinary panics to structured errors (while not relying on it for OOM safety).
5. Add tests at the maximum, maximum+1, `math.MaxInt`, and architecture-specific integer boundaries.

timwhitez:
## Summary

- reject oversized `read.offset` and `read.limit` before opening or allocating
- reject oversized grep context windows and `max_results` before scanning
- cap initial slice capacities independently of the accepted logical limits
- preserve defaults and existing negative-value normalization
- add maximum, maximum+1, and platform `MaxInt` regressions with explicit panic guards

## Validation

- [x] read limit/offset boundaries
- [x] grep before/after/context/max-results boundaries
- [x] no-panic assertions for platform `MaxInt`
- [x] complete Sandbox tests

Closes #20


timwhitez (commented):
Second-pass review completed. Numeric validation runs before dependency lookup, file opening, directory traversal, or model-controlled capacity allocation. Accepted negative values retain the existing normalization/default semantics, and initial capacities are independently capped below the logical output limits. Boundary, MaxInt panic-guard, Sandbox, and full repository tests passed. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `cfa1e8c92bbecc13c6babb4e699c4ba3767b4ab9`.

- `go build ./...` passed

PR #51, opened 2026-08-22
fix(openai): reject truncated SSE streams as incomplete

### LINKED ISSUE #7: [openai] truncated SSE responses are reported as successfully completed
## Summary

Both OpenAI streaming adapters treat a clean HTTP EOF as successful completion even when the protocol's terminal event was never received. A connection that closes after emitting only a prefix therefore produces `StreamDoneEvent` instead of `StreamErrorEvent`, and callers cannot distinguish a complete answer from a truncated one.

Relevant code:

- [Chat Completions recognizes `[DONE]`, but also emits `StreamDoneEvent` when the SSE reader returns nil at ordinary EOF](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/chat.go#L418-L435)
- [Responses parses `response.completed`, but does not record that it was seen](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/responses.go#L824-L865)
- [Responses likewise emits `StreamDoneEvent` after ordinary EOF without requiring `[DONE]` or `response.completed`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/llm/openai/responses.go#L873-L889)

## Reproduction

Use an `httptest.Server` that returns `Content-Type: text/event-stream`, writes one valid text-delta event, flushes it, and then closes the response body without sending a terminal marker.

### Chat Completions

Return a valid `chat.completion.chunk` with a text delta, but omit `data: [DONE]`.

### Responses API

Return a valid `response.output_text.delta`, but omit both `response.completed` and `data: [DONE]`.

Consume the returned event channel. In both cases the observed sequence is a text delta followed by `StreamDoneEvent`; no error is emitted.

## Actual behavior

A transport-level truncation is surfaced as normal completion. The partial text/tool-call arguments can be accepted and persisted as a finished model turn.

## Expected behavior

The adapter must require the provider-specific terminal condition:

- Chat Completions: `[DONE]` (or another explicitly supported, documented terminal event).
- Responses: `response.completed`/another valid terminal response event, with its status interpreted appropriately.

EOF before that condition must emit an incomplete-stream error while preserving already-received deltas as partial output.

## Impact

- Network resets, proxy truncation, or provider failures silently corrupt model output.
- Partial JSON tool arguments may be treated as a complete tool call.
- Agent history and recovery state can record an incomplete assistant response as successful.
- Retry logic is bypassed because no error is surfaced.

The code has explicit handling for terminal markers, but the ordinary-EOF branch unconditionally emits the same success event. There is no `sawDone`/`sawCompleted` state or equivalent validation.

## Suggested fix

Track whether a valid terminal event was observed. After `consumeSSEWithBodyClose` returns nil:

1. emit `StreamDoneEvent` only when the terminal flag is set;
2. otherwise emit a typed `IncompleteStreamError` containing provider/model/endpoint context;
3. preserve the partial deltas for callers that intentionally retain partial output.

Add regression tests for EOF before any terminal marker, including a stream containing a partial tool-call argument sequence.

timwhitez:
## Summary

- require `[DONE]` for Chat Completions streaming
- require `response.completed` or `[DONE]` for Responses streaming
- emit a typed `ProviderError` on ordinary EOF before the provider terminal condition
- preserve already delivered text and tool-call deltas as partial output
- retain valid behavior for explicit Chat and Responses terminal events

## Dependency

This PR is intentionally stacked on #44, whose forwarding layer guarantees cancellation can drain an abandoned producer. Its diff against that branch contains only terminal-condition enforcement and focused regressions.

## Validation

- [x] Chat text EOF regression
- [x] partial tool-argument EOF regression
- [x] Responses EOF regression
- [x] explicit terminal success regressions
- [x] complete OpenAI provider tests

Closes #7


timwhitez (commented):
Second-pass review completed after CI and cleanup. Chat now succeeds only on `[DONE]`; Responses succeeds only on `response.completed` or `[DONE]`. Ordinary EOF yields a provider error while preserving previously delivered text/tool deltas. The explicit-terminal regressions and complete OpenAI/full repository suites passed. Final branch contains no temporary automation. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `4fe5e2619187d3a2b9fd5b02eaa7cfb7d24f18fa`.

- `go build ./...` passed

Transparency: the first full run failed once in the unrelated baseline `TestQueryStreamBackpressureDropsWithoutDeadlock`. SDK `main` and this exact branch each passed that test 20 consecutive times, and the full branch retry passed; no #51 source change was required.

PR #47, opened 2026-08-22
fix(sandbox): reject stale apply_patch targets

timwhitez:
Closed as superseded by #53.

PR #53 is based on #28's final atomic no-clobber source commit, preserves the corrected move primitive, and has completed focused, race, complete Sandbox, and full repository validation. This stale branch predates the atomic correction.

No commits from this PR were merged.

PR #53, opened 2026-08-23
fix(sandbox): reject stale apply_patch targets

### LINKED ISSUE #16: [sandbox] apply_patch overwrites or deletes files changed after planning
## Summary

`apply_patch` reads and transforms target files in memory, then commits the staged actions later. Unlike the `write`, `edit`, and `multiedit` tools, the patch planner does not retain a `FileInfo`/content snapshot and does not verify that the target is still the same file with the same contents before the mutation lands.

At commit time it only re-resolves the pathname and compares the resulting path string. A concurrent writer can replace or modify the file at the same path; that passes revalidation and is silently overwritten. A planned delete can likewise remove a different file created at the same path after planning.

Relevant code:

- [`planUpdate` reads the old content and stores only the resolved path plus new bytes](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_patch.go#L481-L527)
- [`commitAction` only revalidates the path string before `os.Remove`, `os.Rename`, or `writeFilePreserveMode`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_patch.go#L557-L607)
- By contrast, [`write`/`edit` use `writeTargetSnapshot` and compare inode, mode, size, mtime, and full content](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_write.go#L20-L90)

## Deterministic reproduction

This can be reproduced in a package-level regression test using the existing `beforeSandboxPathRevalidate` test hook:

```go
func TestApplyPatchRejectsTargetChangedAfterPlanning(t *testing.T) {
    root := t.TempDir()
    target := filepath.Join(root, "target.txt")
    if err := os.WriteFile(target, []byte("old\n"), 0o644); err != nil {
        t.Fatal(err)
    }
    s, err := New(root)
    if err != nil {
        t.Fatal(err)
    }

    originalHook := beforeSandboxPathRevalidate
    defer func() { beforeSandboxPathRevalidate = originalHook }()

    calls := 0
    beforeSandboxPathRevalidate = func(path string) {
        if filepath.Clean(path) != filepath.Clean(target) {
            return
        }
        calls++
        // planUpdate performs two revalidations while reading. The third is
        // commitAction's final path-only revalidation.
        if calls == 3 {
            if err := os.WriteFile(target, []byte("concurrent writer\n"), 0o644); err != nil {
                panic(err)
            }
        }
    }

    patch := `*** Begin Patch
*** Update File: target.txt
@@ -1,1 +1,1 @@
-old
+patched
*** End Patch`

    _, err = applyPatchToSandbox(s, patch)
    if err == nil {
        t.Fatal("expected stale-target error")
    }
    got, _ := os.ReadFile(target)
    if string(got) != "concurrent writer\n" {
        t.Fatalf("concurrent update was lost: %q", got)
    }
}
```

Current behavior is that `err == nil` and the final content is `patched\n`, proving the concurrent update is lost.

A delete variant is even simpler: replace the planned target with a new regular file at the same path immediately before the delete commit; `revalidatePathForAccess` sees the same resolved string and `os.Remove` deletes the replacement.

## Actual behavior

`apply_patch` reports success while discarding a concurrent update, or deletes/moves a replacement file that was not the object originally planned.

## Expected behavior

Every commit action should fail closed with `workspace_file_changed` when the target identity or contents differ from the planner's snapshot. No stale patch should overwrite/delete a later write.

## Impact

- Silent loss of user or process changes.
- Multi-agent/parallel-tool races can corrupt a workspace.
- A long multi-file patch widens the planning-to-commit window for later actions.
- Behavior is inconsistent with `write`, `edit`, and `multiedit`, which already reject stale targets.

The repository already treats stale write targets as a correctness error in the other mutation tools. The reproduction changes only file contents at the same pathname, so all current path/symlink checks pass by design; the staged bytes then deterministically overwrite the newer data.

## Suggested fix

Store an immutable snapshot for every planned source/destination:

1. Existing targets: inode identity (`os.SameFile`), mode, size, mtime, and preferably the exact bytes/hash used by the planner.
2. Missing add/move destinations: record non-existence and reject if created before commit.
3. Delete/move sources: require the same identity and, where relevant, same content.
4. Reuse the `verifyWriteTargetUnchanged` semantics already implemented by the other tools.

Add regression tests for update, delete, move source, and move destination races.

timwhitez:
## Summary

- retain the first complete on-disk snapshot for every path touched by a patch
- verify identity, metadata, and exact content immediately before writes, deletes, and move-source mutations
- update expected snapshots after successful actions so multi-operation patches do not reject their own writes
- preserve #28's atomic, non-clobbering move primitive
- add deterministic stale update/delete/move-source, composed-update, and destination-race regressions

## Dependency

This replacement PR is based on #28's final source commit `08ad5b900e30a6755782a95ac28947c3f646fa15`. It supersedes stale PR #47, whose base predates the atomic no-clobber correction.

## Validation

- [x] stale update/delete/move-source and composed-update regressions
- [x] final-check destination race regressions retained from #28
- [x] complete Sandbox tests
- [x] final source-only stacked diff review

Validation run: `32638750980`.

Closes #16


timwhitez (commented):
Final stacked review completed after rebasing onto #28's atomic no-clobber move implementation. Every update/delete/move source is compared against the first complete on-disk snapshot immediately before mutation; destination non-existence and atomic creation remain enforced by the parent branch. Expected state is advanced only after successful patch-owned mutations, preserving valid multi-operation patches. Focused tests repeated 20 times, race tests repeated 10 times, complete Sandbox tests, and the full suite passed in run `32638750980`. No merge performed.

timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `6950cdc0dc6650828e93f9342f007588dbe17d7e`.

- `go build ./...` passed

PR #50, opened 2026-08-22
fix(sandbox): defer WebFetch DNS until user confirmation

timwhitez:
Superseded by clean replacement PR #54, based on the final WebFetch socket-pinning source from PR #42.

No commits from this duplicate Draft were merged.

PR #54, opened 2026-08-23
fix(sandbox): defer WebFetch DNS until user confirmation

### LINKED ISSUE #4: [security] webfetch performs attacker-controlled DNS lookups before user confirmation
## Summary

`webfetch` resolves the requested hostname as part of destination validation before it calls the configured `Confirmer`. A request that the user ultimately denies has therefore already caused an outbound DNS query to an attacker-controlled name.

Relevant code:

- [`validateWebfetchDestinationURL` is invoked before confirmation metadata is built](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L52-L65)
- [The confirmation is not requested until later](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L79-L93)
- [Hostname validation calls `LookupIPAddr`](https://github.com/timwhitez/agent-sdk-golang/blob/4f305ce4c134665f2c70c05d4230d8fcc4175935/sdk/tools/sandbox/sandbox_webfetch.go#L158-L203)

## Reproduction

1. Configure a confirmer that always returns `false`.
2. Invoke `webfetch` with a unique hostname under an authoritative DNS zone you control, for example `<secret-or-unique-token>.attacker.example`.
3. Observe the authoritative DNS server.

The lookup arrives even though the tool result is denied and no HTTP request should have been authorized.

A unit test can override `webfetchLookupIPAddrs` with a function that increments a counter, use a denying confirmer, invoke the tool, and assert that the counter remains zero. It currently becomes one.

## Actual behavior

Network activity occurs before the user/security policy approves the network tool call.

## Expected behavior

Before approval, `webfetch` should perform only local/syntactic validation. DNS resolution and all other network I/O must start only after confirmation succeeds.

## Impact

- A model-controlled URL can exfiltrate data through DNS labels even when the user denies the request.
- Denied calls still disclose that the host/application attempted to resolve a name.
- This violates the SDK's confirmation boundary for network actions.

The function call order is unconditional: destination validation executes first, and non-literal hosts always call the resolver. The confirmer is reached only after that lookup returns successfully.

## Suggested fix

Split validation into two phases:

1. Before confirmation: parse the URL, validate the scheme/host syntax, and reject literal denied IPs without network I/O.
2. After confirmation: resolve, classify, and pin the accepted addresses for the actual dial.

Add a regression test with a denying confirmer and an instrumented resolver that must not be called.

timwhitez:
## Summary

- perform only local URL and literal-IP checks before confirmation
- defer hostname resolution to #42's socket-bound validator
- guarantee a denied hostname request performs zero DNS lookups and zero socket dials
- keep blocked literal destinations fail-closed without prompting
- retain exact-socket DNS classification after approval

## Dependency

This replacement PR is based on #42's corrected package-owned WebFetch transport. It supersedes closed PR #50, whose branch contained temporary patch automation.

## Local validation

Audited locally at exact head `bf2b51f33d9abffa861f9cdf28b187e24745ec5d` with its exact #42 parent, using Go 1.24.5 on 2026-08-24; no GitHub Actions were invoked:

- [x] `gofmt` clean
- [x] `go vet ./...`
- [x] focused Sandbox tests
- [x] focused Sandbox race tests
- [x] `go build ./...`
- [x] final stacked source-only diff review

Closes #4


timwhitez:
Local closed-loop re-audit (2026-08-24, official Go 1.24.5) completed at exact head `bf2b51f33d9abffa861f9cdf28b187e24745ec5d`.

- `go build ./...` passed

PR #62, opened 2026-08-27
fix(openai): correlate Responses tool outputs by call_id

### LINKED ISSUE #55: [openai responses] correlate function outputs with call_id instead of item id
## Summary

Responses function-call output items carry two identifiers: an output item `id` (normally `fc_...`) and the function correlation `call_id` (normally `call_...`). The SDK currently prefers `item.id` in both buffered and streaming parsing, then exposes it as `llm.ToolCall.ID`. The next request serializes that value as `function_call_output.call_id`, so official-shaped responses are correlated with the wrong identifier.

Existing tests use synthetic `id: "call_*"` fixtures and do not cover the official dual-ID shape.

## Affected paths

- `sdk/llm/openai/responses.go`: `response.output_item.added`, `response.output_item.done`, argument-delta correlation, and `parseResponses`
- Responses stream/non-stream tests

## Expected behavior

- `llm.ToolCall.ID` uses the provider `call_id` that tool outputs must reference.
- Output item `id` remains available only for correlating streaming events that use `item_id`.
- Interleaved/parallel stream calls remain one-to-one.

## Acceptance criteria

- Buffered and streaming official-shaped fixtures include distinct `id=fc_*` and `call_id=call_*`.
- The follow-up request emits `function_call_output.call_id=call_*`, never `fc_*`.
- Argument deltas keyed by `item_id` still attach to the correct call.
- Full SDK tests and race tests pass.

timwhitez:
## Summary

- keep Responses output item IDs separate from function `call_id` values
- expose `call_id` as the provider-neutral tool-call ID used by follow-up outputs
- bind both identifiers to one streaming index so interleaved argument deltas remain correlated
- retain the single-ID fallback used by compatible gateways

## Validation

- strict buffered official dual-ID fixture
- strict streaming `item_id=fc_*` / `call_id=call_*` fixture
- strict two-request fixture asserting `function_call_output.call_id=call_*`
- `go test -count=1 ./sdk/llm/openai`
- `go test -race -count=1 ./sdk/llm/openai`
- `go test -count=1 ./...`
- `go vet ./sdk/llm/openai`
- `git diff --check`

Closes #55

### DIFF
--- a/sdk/llm/openai/responses.go
+++ b/sdk/llm/openai/responses.go
@@ -694,6 +694,18 @@ func (c *ResponsesClient) InvokeStream(ctx context.Context, req llm.InvokeReques
 				emitToolIdentity(idx, itemID, "")
 				return nil
 			}
+			recordStreamItemID := func(idx int, itemID string) error {
+				itemID = strings.TrimSpace(itemID)
+				if itemID == "" {
+					return nil
+				}
+				st := ensureToolState(idx)
+				if st.itemID != "" && st.itemID != itemID {
+					return fmt.Errorf("openai responses stream: tool index %d changed item_id from %q to %q", idx, st.itemID, itemID)
+				}
+				st.itemID = itemID
+				return nil
+			}
 
 			stopReason := ""
 			sawCompleted := false
@@ -881,6 +893,9 @@ func (c *ResponsesClient) InvokeStream(ctx context.Context, req llm.InvokeReques
 						if indexErr != nil {
 							return indexErr
 						}
+						if itemErr := recordStreamItemID(idx, itemID); itemErr != nil {
+							return itemErr
+						}
 						st := ensureToolState(idx)
 						if promoteErr := promoteCompatibilityItemID(idx, itemID); promoteErr != nil {
 							return promoteErr
@@ -909,6 +924,9 @@ func (c *ResponsesClient) InvokeStream(ctx context.Context, req llm.InvokeReques
 						idx, indexErr := bindToolIndex(itemID, "", autoKey, idxHint, hasIdxHint)
 						if indexErr != nil {
 							return indexErr
+						}
+						if itemErr := recordStreamItemID(idx, itemID); itemErr != nil {
+							return itemErr
 						}
 						if promoteErr := promoteCompatibilityItemID(idx, itemID); promoteErr != nil {
 							return promoteErr
--- a/sdk/llm/openai/responses.go
+++ b/sdk/llm/openai/responses.go
@@ -497,7 +497,8 @@ func (c *ResponsesClient) InvokeStream(ctx context.Context, req llm.InvokeReques
 				return
 			}
 
-			idToIndex := map[string]int{}
+			itemIDToIndex := map[string]int{}
+			callIDToIndex := map[string]int{}
 			autoToIndex := map[string]int{}
 			usedIndices := map[int]struct{}{}
 			nextIndex := 0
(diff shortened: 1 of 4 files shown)


timwhitez:
INDEPENDENT REVIEW BLOCK

Official contract checked:
- A function_call output item has two distinct identities: id is the output-item identity; call_id is the function correlation identity used by function_call_output.
- response.function_call_arguments.delta is correlated by item_id plus output_index.
- Sources: https://developers.openai.com/api/reference/resources/responses/streaming-events and https://developers.openai.com/api/reference/resources/responses/subresources/input_items/methods/list

Blocking defect 1 — late or out-of-order call_id cannot replace the item-ID fallback:
1. If response.function_call_arguments.delta arrives before output_item.added, responses.go:849-854 emits item_id (for example fc_A) as StreamToolCallDeltaEvent.ID because streamedToolState.callID is still empty.
2. A later output_item.added/done can discover call_id=call_A and emit/set it, but sdk/agent/agent.go:1584-1588 only records an ID when the accumulator ID is empty.
3. The finalized ToolCall therefore remains fc_A. The next request serializes function_call_output.call_id=fc_A, reproducing Issue #55 for this supported compatibility/out-of-order path.
The same failure occurs when output_item.added initially contains only id=fc_A and output_item.done later supplies call_id=call_A: the fallback marks hasID=true, so the real call_id is not emitted as a correction.

Blocking defect 2 — item IDs and call IDs share one alias namespace:
responses.go:587-621 stores both domains in idToIndex. If one parallel call's call_id equals another call's item id, or late aliases conflict, getIndexForIDs selects the first mapping and silently rewrites the other alias. Two distinct calls can then merge into one index. The official schema defines separate identities and does not establish a cross-domain prefix/disjointness contract.

Required before approval:
- Keep item_id-to-index and call_id-to-index mappings in separate namespaces, and reject or safely resolve conflicting aliases instead of silently merging.
- Do not commit item_id as the provider-neutral ToolCall.ID until call_id is known to be unavailable at a terminal boundary, or add a safe, explicit ID-upgrade contract through the accumulator.
- Add tests for delta-before-added, call_id appearing only at done, two interleaved official dual-ID calls, cross-domain alias conflict, and a streaming completion followed by a request that asserts function_call_output.call_id.

Validation at 36ca310:
- Focused call-id tests: PASS.
- go test -count=1 ./sdk/llm/openai: PASS.
- go test -race -count=1 ./sdk/llm/openai: PASS.
- go test -count=1 ./... in golang:1.22 with Docker init: PASS.
- go test -race -count=1 ./... in golang:1.22 with Docker init: PASS.
- go vet ./...: PASS.
- git diff --check origin/main...HEAD: PASS.
- GitHub check-runs: none reported.

The first full-suite run without a container init failed only TestBashTool_KillsProcessGroupOnTimeout because PID 1 did not reap the killed orphan. The same focused test failed without init and passed with Docker --init; the full suite then passed with init. This is environment evidence, not attributed to this PR.

No files were modified and no provider credential was read or printed.

timwhitez:
Addressed the independent review blocker in commit `HEAD`:

- split `item_id` and `call_id` into separate streaming correlation namespaces
- reject contradictory same-domain/output-index mappings instead of silently merging calls
- buffer argument deltas until a late authoritative `call_id` arrives, so `fc_*` never becomes the downstream tool-call ID
- retain a bounded compatibility promotion only when a named single-ID gateway never supplied either identity
- add out-of-order SSE, cross-namespace collision, and streamed two-request `function_call_output.call_id` regressions

Revalidation: OpenAI package, package race, full SDK, full vet, and diff-check all pass. Requesting independent re-review.

timwhitez:
INDEPENDENT RE-REVIEW BLOCK — head 5f827d223b5e6fc0631d4bc35ebea24d7bef876a

The previous three functional blockers are materially improved:
- item_id and call_id now use separate maps;
- argument deltas received before the identity are buffered until call_id arrives;
- the new cross-namespace same-string and streamed two-request regressions pass.

One fail-closed defect remains.

Contradictory item_id values can still be silently attached to an already-bound output index:

1. Receive response.output_item.added with output_index=0 and item {id:fc_a, call_id:call_a, name:read}. This records streamedToolState.itemID=fc_a and callID=call_a.
2. Receive response.function_call_arguments.delta with output_index=0, item_id=fc_b, and arguments for another item.
3. At responses.go:880, bindToolIndex sees that fc_b is new but output:0 already maps to index 0. Lines 594-601 then add fc_b as another alias of index 0 instead of rejecting the mismatch.
4. The delta path at responses.go:884-891 never compares the incoming item_id with streamedToolState.itemID. Because call_a is already present, it emits fc_b's arguments as StreamToolCallDeltaEvent{ID:call_a}.

Result: no StreamErrorEvent is produced, and arguments from a contradictory output item are assigned to call_a. This violates the official contract that a function-call argument delta's item_id identifies the output item at output_index, and it is not fail closed.

Official references:
- https://developers.openai.com/api/reference/resources/responses/streaming-events
- https://developers.openai.com/api/reference/resources/responses/subresources/input_items/methods/list

Required before approval:
- Maintain inverse identity for each bound stream index, or set/check streamedToolState.itemID on every delta.
- Reject a non-empty incoming item_id that differs from the index's established item_id before storing the alias or emitting arguments.
- Add a regression with fc_a/call_a at output_index 0 followed by a delta carrying item_id=fc_b at output_index 0; assert one StreamErrorEvent and no argument delta attributed to call_a.
- Also cover the symmetric delta-first case: fc_a delta first, then an identity item fc_b at the same output index.

Read-only validation at 5f827d2:
- Five focused late-ID, namespace-collision, and streamed-round-trip tests: PASS.
- go test -count=1 ./sdk/llm/openai: PASS.
- go test -race -count=1 ./sdk/llm/openai: PASS.
- go test -count=1 ./...: PASS.
- go test -race -count=1 ./...: PASS.
- go vet ./...: PASS.
- git diff --check: PASS.
- GitHub check-runs: none reported.

Tests used golang:1.22 with Docker --init and a read-only repository mount. No files were modified and no provider credential was read or printed.

timwhitez:
Resolved the second independent-review blocker: each streaming tool index now records one immutable `item_id`; both identity-first and delta-first conflicting IDs fail with `StreamErrorEvent` before conflicting arguments are emitted. Added regressions for both event orders. OpenAI package, race, full SDK, full vet, and diff-check pass. Requesting re-review of head.

timwhitez:
INDEPENDENT REVIEW PASS — head ff643665cf507c6ea7c1db8b0fb255244cd1b27e

Official Responses contract rechecked: response.function_call_arguments.delta identifies one output item through item_id and output_index.
Source: https://developers.openai.com/api/reference/resources/responses/streaming-events

The prior BLOCK is resolved:
- recordStreamItemID makes the non-empty item_id immutable for each streaming tool index.
- Both delta and done paths call it before compatibility promotion, buffering, or emitToolArgs.
- Identity-first conflict (fc_a identity, then fc_b delta at the same index) produces StreamErrorEvent before the conflicting arguments are emitted.
- Delta-first conflict (fc_a delta, then fc_b delta at the same index) also produces StreamErrorEvent with no conflicting argument event.
- The earlier late-call_id buffering, separate call/item namespaces, cross-namespace same-string collision, output-index correlation, and streamed two-request function_call_output.call_id behavior remain intact.

Read-only validation at ff64366:
- Seven focused conflict/late-ID/namespace/round-trip tests: PASS.
- go test -count=1 ./sdk/llm/openai: PASS.
- go test -race -count=1 ./sdk/llm/openai: PASS.
- go test -count=1 ./...: PASS.
- go test -race -count=1 ./...: PASS.
- go vet ./...: PASS.
- git diff --check: PASS.
- GitHub check-runs: none reported; PR is mergeable and merge state is clean.

Tests ran with golang:1.22, Docker --init, and a read-only repository mount. No files were modified and no provider credential was read or printed.

PR #63, opened 2026-08-27
fix(agent): honor zero stream idle recoveries

### LINKED ISSUE #57: [agent] StreamIdleMaxRecoveries zero does not disable automatic retries
## Summary

`Config.StreamIdleMaxRecoveries` documents `0` as disabling auto-recovery and negative values as selecting the default. `New` preserves zero, but the idle-timeout branch starts with the package default and only overrides it when the configured value is greater than zero. A configured zero therefore still retries twice.

This can repeat provider calls and billing after the caller explicitly disabled recovery, while also injecting an internal recovery message.

## Acceptance criteria

- `0`: first idle timeout terminates with the typed error and no recovery request/message.
- positive N: exactly N recoveries are allowed.
- negative: package default remains in effect.
- streaming usage and partial-output behavior remain unchanged.

timwhitez:
## Summary

- preserve the documented three-way recovery semantics: negative selects the default, zero disables, positive values set an explicit cap
- stop falling back to the package default after `New` has normalized configuration
- update recovery tests to state their intended configuration explicitly
- add a zero-recovery regression asserting one provider call and no internal retry message

## Validation

- focused idle stream tests
- `go test -count=1 ./sdk/agent`
- focused `go test -race`
- `go test -count=1 ./...`
- `go vet ./sdk/agent`
- `git diff --check`

Closes #57

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -728,7 +728,7 @@ func (a *Agent) QueryStreamWithSteering(ctx context.Context, input llm.Content, 
 				var idleErr *llm.StreamIdleTimeoutError
 				if errors.As(err, &idleErr) {
 					maxRecov := agentStreamIdleMaxRecoveries
-					if a != nil && a.streamIdleMaxRecov > 0 {
+					if a != nil {
 						maxRecov = a.streamIdleMaxRecov
 					}
 					if streamIdleRecoveries < maxRecov {
--- a/sdk/agent/agent_stream_idle_test.go
+++ b/sdk/agent/agent_stream_idle_test.go
@@ -72,7 +72,7 @@ func TestQueryStreamAutoRecoversFromIdleStream(t *testing.T) {
 	})
 
 	model := &streamIdleRecoveryModel{}
-	ag, err := New(Config{LLM: model})
+	ag, err := New(Config{LLM: model, StreamIdleMaxRecoveries: -1})
 	if err != nil {
 		t.Fatalf("new agent: %v", err)
 	}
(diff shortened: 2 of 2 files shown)


timwhitez (commented):
INDEPENDENT REVIEW PASS

Reviewed Issue #57 and PR #63 at head `63778aecd7e9555cfdf20b7a15f548386a9cb755` without modifying the worktree.

Acceptance/semantic verification:
- `Config.StreamIdleMaxRecoveries` documents the three states accurately: negative selects the package default, zero disables recovery, positive sets an explicit cap.
- `New` normalizes only negative values to the package default, preserves zero and positive values, and stores the normalized value on `Agent`.
- The idle-timeout branch now consumes that stored value even when it is zero. The `< maxRecov` counter permits exactly N recoveries for positive N; zero immediately reaches the terminal timeout path.
- The terminal timeout path still appends partial assistant content and emits any partial provider usage exactly once. Recovery paths retain the existing partial-history and per-attempt usage behavior.
- The zero regression proves exactly one provider call and no `sdk_internal_stream_idle_recovery` message, so disabling recovery cannot create an extra SDK retry/provider-billing opportunity.

Validation (official Go 1.22.12 container, repository bind-mounted read-only, no provider credentials supplied):
- focused four idle-stream tests: PASS
- `go test -count=1 ./sdk/agent`: PASS
- `go test -race -count=1 ./...`: PASS
- `go test -count=1 ./...`: PASS
- `go vet ./...`: PASS
- `git diff --check origin/main...HEAD`: PASS
- post-test worktree status: clean

GitHub reports the PR MERGEABLE/CLEAN and currently reports no status checks for this branch. I found no correctness, regression, history, usage-accounting, or billing blocker.

Formal approval is unavailable because the authenticated GitHub account is also the PR author; this comment records the independent sub-agent review result.

PR #67, opened 2026-08-27
fix(openai): keep bearer credentials on the configured origin

### LINKED ISSUE #61: [security] OpenAI clients may forward bearer credentials across redirects
## Summary

OpenAI Chat and Responses clients use the default `http.Client` redirect behavior. Their buffered and streaming requests attach `Authorization: Bearer ...`, but there is no same-origin or HTTPS-downgrade redirect policy comparable to the Anthropic client.

A 30x response can therefore move a credential-bearing request to another origin. In particular, same-host/different-port redirects retain sensitive headers under Go's host-based forwarding rules.

## Acceptance criteria

- Chat and Responses, buffered and streaming, share a redirect-safe client policy.
- Credential-bearing requests follow only same-origin redirects; HTTPS-to-HTTP is rejected.
- A caller-provided `CheckRedirect` is composed for allowed redirects and cannot mutate the request onto a disallowed origin.
- Tests cover cross-host, same-host/different-port, scheme downgrade, allowed same-origin, and stream client cloning.
- Error messages identify the rejected redirect without exposing credentials.

timwhitez:
## Summary

- install one mandatory same-origin redirect policy for OpenAI Chat and Responses
- preserve the policy in buffered and streaming clients
- define origin by scheme, hostname, and effective port; reject HTTPS downgrade
- compose caller redirect callbacks and recheck their mutated target
- sanitize `url.Error` diagnostics to origin-only labels so malicious Location userinfo/path/query cannot leak

## Validation

- Chat/Responses × buffered/streaming cross-origin rejection with zero destination requests
- Chat/Responses × buffered/streaming same-origin success with bearer preservation
- same-host/different-port and HTTPS downgrade tests
- caller callback recheck and `ErrUseLastResponse` tests
- credential-bearing Location diagnostic redaction
- `go test -count=1 ./sdk/llm/openai`
- `go test -race -count=1 ./sdk/llm/openai`
- `go test -count=1 ./...`
- `go vet ./...`
- `git diff --check`

Closes #61

### DIFF
--- a/agent_docs/providers.md
+++ b/agent_docs/providers.md
@@ -100,6 +100,15 @@ stream normalization, and response metadata behavior.
 - Anthropic populates IDs in both sync and streaming paths (`sdk/llm/anthropic/client.go:718`, `sdk/llm/anthropic/client.go:1219`)
 - OpenAI Chat completions currently do not expose response IDs in parsed completion objects (`sdk/llm/openai/chat.go:904`)
 
+## OpenAI redirect credential boundary
+
+OpenAI Chat and Responses clients, in buffered and streaming modes, follow
+only same-origin redirects. Origin includes scheme, host, and effective port;
+HTTPS-to-HTTP downgrades are rejected before a redirected request is sent.
+Caller-provided redirect callbacks run only for an initially allowed target,
+and the target URL is checked again after the callback returns so it cannot
+move bearer credentials onto another origin.
+
 ## Usage and Optional Cost Calculation
 - Providers populate a versioned `Usage` contract; downstream cost calculation is optional (`sdk/llm/types.go`, `sdk/llm/usage.go`, `sdk/tokens/cost.go:80`).
 - Under `prompt_tokens_semantics=total_input_v1`, `PromptTokens` is the complete effective input size. `PromptCachedTokens`, `PromptCacheCreationTokens`, `PromptImageTokens`, and `PromptUncachedTokens` are breakdown fields and consumers must not add them to `PromptTokens` again.
--- a/agent_docs/providers.md
+++ b/agent_docs/providers.md
@@ -107,7 +107,9 @@ only same-origin redirects. Origin includes scheme, host, and effective port;
 HTTPS-to-HTTP downgrades are rejected before a redirected request is sent.
 Caller-provided redirect callbacks run only for an initially allowed target,
 and the target URL is checked again after the callback returns so it cannot
-move bearer credentials onto another origin.
+move bearer credentials onto another origin. Redirect-policy and malformed
+Location failures are fail-fast and their diagnostics never echo URL
+userinfo, path, or query data.
 
 ## Usage and Optional Cost Calculation
 - Providers populate a versioned `Usage` contract; downstream cost calculation is optional (`sdk/llm/types.go`, `sdk/llm/usage.go`, `sdk/tokens/cost.go:80`).
--- a/sdk/llm/openai/chat.go
+++ b/sdk/llm/openai/chat.go
@@ -123,6 +123,7 @@ func (c *ChatClient) Invoke(ctx context.Context, req llm.InvokeRequest) (*llm.Co
 		}
 
 		resp, err := client.Do(httpReq)
+		err = sanitizeOpenAIHTTPError(err)
 		if err == nil {
 			data, readErr := readResponseBodyLimited(resp.Body, endpoint)
 			if readErr != nil {
@@ -256,6 +257,7 @@ func (c *ChatClient) InvokeStream(ctx context.Context, req llm.InvokeRequest) (<
 			}
 
 			resp, err := client.Do(httpReq)
+			err = sanitizeOpenAIHTTPError(err)
 			if err != nil {
 				if ctxErr := ctx.Err(); ctxErr != nil {
 					out <- llm.StreamErrorEvent{Err: ctxErr}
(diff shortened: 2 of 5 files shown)


timwhitez:
## Independent review: **BLOCK**

Reviewed exact head `f8044c7ce4812018a50380602cb12a3e2515da29`. Do not merge yet.

### Blocking findings

1. **Malformed `Location` values still leak userinfo, path, and query data in diagnostics.**

   [`sanitizeOpenAIHTTPError`](https://github.com/timwhitez/agent-sdk-golang/blob/f8044c7ce4812018a50380602cb12a3e2515da29/sdk/llm/openai/redirect.go#L63-L76) replaces only the outer `url.Error.URL` and preserves `urlErr.Err`. In Go's `net/http`, a redirect whose `Location` cannot be parsed fails before `CheckRedirect` and places the raw header in that inner error.

   Disposable adversarial test: a 307 with `Location: http://audit-user:audit-password@%zz/private?token=audit-query-secret` produced an error containing `audit-user`, `audit-password`, `/private`, and `audit-query-secret`. This violates the issue's credential-safe diagnostic acceptance criterion. Add Chat/Responses × buffered/streaming malformed-`Location` redaction coverage.

2. **Redirect-policy failures can be attacker-controlled into the network retry path.**

   The safe redirect error includes the destination origin, while [`isRetryableNetErr`](https://github.com/timwhitez/agent-sdk-golang/blob/f8044c7ce4812018a50380602cb12a3e2515da29/sdk/llm/openai/chat.go#L933-L945) classifies errors by substrings such as `connection`, `timeout`, and `tls`. A cross-origin `Location` using host `connection.invalid` therefore becomes retryable. With `MaxRetries: 3`, a disposable probe observed 3 POSTs to the redirect source instead of one fail-fast POST. The default policy can amplify this further with exponential backoff.

   Use a typed non-retryable redirect-policy error (including malformed redirect parsing) and assert one source request for both clients in buffered and streaming modes.

### Confirmed behavior

- Chat and Responses, buffered and streaming, reject valid cross-origin redirects before the destination receives a request.
- Same-origin redirects preserve the bearer; effective-port comparison rejects same-host/different-port, and HTTPS downgrade is rejected.
- Caller `CheckRedirect` is composed and its URL mutation is rechecked; `ErrUseLastResponse` is preserved.
- The base client is cloned, and the streaming timeout clone retains the mandatory redirect callback. I found no normal-path double wrapping or timeout-clone regression.

### Validation on the unmodified PR head

- PASS: focused redirect tests
- PASS: `go test -count=1 ./sdk/llm/openai`
- PASS: `go test -race -count=1 ./sdk/llm/openai`
- PASS: `go test -count=1 ./...`
- PASS: `go vet ./...`
- PASS: `git diff --check origin/main...HEAD`
- BLOCK reproduced: the two disposable adversarial tests above

No real provider credential was used.

timwhitez:
Addressed both review blockers in ed6f2e3. Redirect-policy errors now carry a typed fail-fast decision before message heuristics, so hostnames such as connection.invalid cannot trigger retries. URL-error sanitization discards untrusted nested text (including malformed Location values) while preserving safe timeout/network retry semantics. Added Chat/Responses × buffered/streaming tests for malformed Location redaction and one-attempt behavior, retry-keyword policy errors, same-origin embedded URL credentials, and sanitized timeout/network classification. Focused, OpenAI package, race, full ./..., vet, and diff-check pass.

timwhitez:
## Independent re-review: **PASS**

Reviewed exact head `ed6f2e32b0ffdc84f3512729f72084d94fe09a99`. The two prior blockers are resolved; I found no new merge blocker in this scope.

### Prior blockers closed

1. **Malformed `Location` diagnostics are fully redacted and fail fast.** The sanitizer no longer preserves the raw nested `url.Error.Err`. Chat/Responses × buffered/streaming tests with `MaxRetries: 3` each made exactly one source request and exposed none of the injected userinfo, password, host keyword, path, query, or sentinel values.

2. **Redirect-policy errors cannot be spoofed into retries by diagnostic text.** Typed retry decisions take precedence over the legacy string fallback. Independent probes covered target hostnames containing each of `connection`, `tls`, and `timeout` across Chat/Responses × buffered/streaming: every case made one source request, zero destination requests, and remained non-retryable.

### Additional boundary verification

- Same-origin redirects still preserve the bearer and succeed in all four client modes.
- Same-origin redirect URL userinfo is rejected before transmission. A caller callback that injected same-origin userinfo/path/query was also rechecked: source/destination/callback counts were exactly 1/0/1 in all four modes, with redacted diagnostics.
- Cross-origin, same-host/different-port, and HTTPS downgrade remain blocked before the destination request.
- Caller callback composition, post-callback URL recheck, `ErrUseLastResponse`, base-client immutability, and streaming-client policy retention all pass.
- Real connection-refused errors remained retryable and produced exactly 3 configured round trips in all four modes.
- Real buffered `http.Client.Timeout` errors retained `Timeout()==true` and the configured 3-attempt retry behavior for both clients.
- Streaming still removes the client-wide timeout; a real context deadline produced one source request and terminal `ctx.Err()==context.DeadlineExceeded`, with no retry.

### Validation on the unmodified PR head

- PASS: focused redirect/security suite
- PASS: independent disposable adversarial probes described above
- PASS: `go test -count=1 ./sdk/llm/openai`
- PASS: `go test -race -count=1 ./sdk/llm/openai`
- PASS: `go test -count=1 ./...`
- PASS: `go vet ./...`
- PASS: `git diff --check origin/main...HEAD`

The disposable probes were removed and the detached review checkout is clean. No real provider credential was used. Review gate: **PASS**.

Issue #76, opened 2026-08-28
Wine can report successful symlink creation without an observable link


timwhitez:
## Summary

After Issue #75 makes the Windows sandbox test package compilable, the real Wine 9 / Windows 10.0.19043 runtime gate deterministically shows that `os.Symlink(...)=nil` is not sufficient evidence that either directory or file symlink fixtures are observable. Thirteen existing security tests can consequently fail or falsely pass against missing or ordinary paths instead of exercising the intended symlink behavior at SDK base `618501028a89a7475bc1d7c295ee108ce0fad723`.

## Reproduction

```text
GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c -o sandbox.test.exe ./sdk/tools/sandbox
wine sandbox.test.exe -test.count=1
```

Affected Windows-reachable tests are:

- `TestSandboxAllowExternalDirectoryResolvesSymlinks`, `TestSandboxResolveBlocksSymlinkEscape`, and the currently false-green `TestSandboxResolveAllowsSymlinkInsideRoot`
- six read/write/edit/ls/glob/grep directory-symlink swap-after-resolve tests
- two glob/grep file-symlink escape tests
- two glob/grep file-symlink swap tests

Representative failures are a missing link immediately after successful creation, failure to remove the supposed link in a swap hook, and absent symlink-warning metadata because the scan saw no link. `TestSandboxResolveAllowsSymlinkInsideRoot` passes without ever exercising an observable link, so a green result is not authoritative either.

## Root cause

The tests treat `os.Symlink(...)=nil` as proof that a usable symlink was created. Under this Wine environment, symlink creation can return success while `Lstat` cannot observe the link. Security assertions then run against an invalid fixture and provide neither a valid pass nor a valid failure.

This failure is unrelated to Issue #75: that issue only moves a POSIX process-group test behind the canonical Unix build constraint and does not change symlink or sandbox production code.

## Expected fix

- Add a test-only helper that requires successful creation, `Lstat` visibility with `ModeSymlink`, and `EvalSymlinks` resolution to the expected target; skip with a precise capability reason otherwise.
- Use the helper across all thirteen Windows-reachable tests above, including setup/probe creation, without hiding genuine post-capability swap errors.
- Preserve the real symlink resolution, escape, warning-metadata, and TOCTOU assertions on capable filesystems.
- Keep POSIX-only Bash swap coverage outside this narrow portability fix.
[...]

PR #79, opened 2026-08-28
Make sandbox mode-preservation tests portable on Windows

### LINKED ISSUE #78: Windows sandbox mode-preservation tests assert unsupported Unix executable bits
## Summary

At exact SDK base `618501028a89a7475bc1d7c295ee108ce0fad723`, a real Wine 9 / Windows 10.0.19043 sandbox test run deterministically fails both mode-preservation tests even though the implementation preserves the mode observable through the Windows Go API.

## Reproduction

```text
GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c -o sandbox.test.exe ./sdk/tools/sandbox
wine sandbox.test.exe -test.run '^(TestWriteToolPreservesMode|TestApplyUpdateFilePreservesMode)$' -test.count=1

--- FAIL: TestWriteToolPreservesMode
    expected mode 0755, got 666
--- FAIL: TestApplyUpdateFilePreservesMode
    expected mode 0755, got 666
```

## Root cause

Both tests create a file with mode `0755`, deliberately skip the precondition/capability check when `runtime.GOOS == windows`, and then unconditionally demand `0755` after the operation. Go on Windows exposes only the portable read-only permission distinction; executable bits are not a supported filesystem contract. The production helper reads the observable pre-operation mode and applies that same mode to the replacement, so an unconditional Unix-mode assertion cannot validate preservation on Windows.

## Expected fix

- Record the observable pre-operation permission state and assert that it is unchanged after write/apply-update.
- Retain the exact `0755` assertion on filesystems that demonstrably support it.
- On Windows, test the supported read-only/writable distinction rather than Unix executable bits, or skip only the unsupported sub-assertion with a precise reason.
- Verify Linux focused/race, Windows cross-vet/test-binary compilation, and real Wine execution.
- Do not use GitHub Actions.

## Scope

This is separate from Issue #76: no symlink behavior is involved. No provider credentials are involved.

timwhitez:
Fixes #78.

## What changed

- Capture each fixture's mode as actually exposed by the target filesystem before the write/update.
- Preserve the exact `0755` assertion on filesystems that expose it.
- On Windows, keep the tests active and validate preservation of the supported writable/read-only mode projection instead of demanding Unix executable bits.
- Change tests only; sandbox production behavior is unchanged.

## Exact revision

- Base: `47b9bd05b13086d4255244952dd163a8614417fe`.
- Head: `94b3eb131ebcb3749d95808abb5549dac760ee64`.
- Head tree and clean merge-tree: `b601b9e3c4119c013e845a30d8a023d3c23c0d76`.

## Direct local evidence

- Linux owning tests: normal count 100 PASS; race count 20 PASS.
- Linux all packages: uncached normal 13/13 PASS; full race 13/13 PASS; vet/build/module verification PASS.
- Windows/amd64: all-package vet PASS, all-package build PASS, sandbox test-binary compilation PASS.
- Real Wine 9 / Windows 10.0.19043: both owning tests count 100 PASS.
- Real Wine full sandbox excluding the separately tracked symlink-capability scope in #76: PASS.
- A diagnostic readonly target remained `0444` after the expected Windows `Access denied`, confirming that this PR must not change production write semantics.

No GitHub Actions were used or queried. Independent exact-head review is required before merge.

### DIFF
--- a/sdk/tools/sandbox/sandbox_test.go
+++ b/sdk/tools/sandbox/sandbox_test.go
@@ -826,13 +826,7 @@ func TestWriteToolPreservesMode(t *testing.T) {
 	if err := os.WriteFile(path, []byte("echo hi\n"), 0o755); err != nil {
 		t.Fatalf("write: %v", err)
 	}
-	if runtime.GOOS != "windows" {
-		if st, err := os.Stat(path); err != nil {
-			t.Fatalf("stat: %v", err)
-		} else if st.Mode().Perm() != 0o755 {
-			t.Skipf("filesystem does not preserve exec perms (got %o)", st.Mode().Perm())
-		}
-	}
+	wantMode := observableModeForPreservationTest(t, path, 0o755)
 	deps := tools.NewContainer()
 	tools.Provide(deps, Key, func(context.Context) (*Sandbox, error) { return s, nil })
 	tools.Provide(deps, ConfirmKey, func(context.Context) (Confirmer, error) { return allowConfirmer{}, nil })
@@ -845,9 +839,32 @@ func TestWriteToolPreservesMode(t *testing.T) {
 	if err != nil {
 		t.Fatalf("stat: %v", err)
 	}
-	if st.Mode().Perm() != 0o755 {
-		t.Fatalf("expected mode 0755, got %o", st.Mode().Perm())
+	if st.Mode().Perm() != wantMode {
+		t.Fatalf("expected observable mode %04o to be preserved, got %04o", wantMode, st.Mode().Perm())
 	}
+}
+
+func observableModeForPreservationTest(t *testing.T, path string, requested os.FileMode) os.FileMode {
+	t.Helper()
+	st, err := os.Stat(path)
+	if err != nil {
+		t.Fatalf("stat mode fixture: %v", err)
+	}
+	observed := st.Mode().Perm()
+	requested = requested.Perm()
+	if observed == requested {
+		return observed
+	}
+	if runtime.GOOS != "windows" {
+		t.Skipf("filesystem does not preserve requested mode %04o (observed %04o)", requested, observed)
+	}
+	// Windows projects the writable/read-only file attribute into Go mode
+	// bits; executable bits are not part of its filesystem contract. Keep the
+	// test active and verify that the writable projection itself is preserved.
+	if requested&0o200 != 0 && observed&0o200 == 0 {
+		t.Fatalf("writable fixture became read-only: requested %04o, observed %04o", requested, observed)
+	}
+	return observed
 }
 
 func TestWriteTool_DiffPreviewLimitsLargeExistingFile(t *testing.T) {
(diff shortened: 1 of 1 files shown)


timwhitez:
FINAL REVIEW: PASS

Independent read-only review of PR #79 / Issue #78 at exact head `94b3eb131ebcb3749d95808abb5549dac760ee64`, exact base/current main `47b9bd05b13086d4255244952dd163a8614417fe`, and exact candidate tree/clean merge-tree `b601b9e3c4119c013e845a30d8a023d3c23c0d76`.

The test now uses the correct platform authority without weakening production coverage.

- Go 1.24.5 maps a writable regular Windows file to observable mode `0666` and a read-only file to `0444`; Unix executable bits are not represented by that API. [The helper captures the actual fixture mode before the operation](https://github.com/timwhitez/agent-sdk-golang/blob/94b3eb131ebcb3749d95808abb5549dac760ee64/sdk/tools/sandbox/sandbox_test.go#L847-L868), rejects an unexpectedly non-writable Windows fixture, and requires the exact same observable projection afterward.
- On Unix, an observed mode different from requested `0755` retains the prior precise capability skip. On this Linux filesystem both owning tests emitted JSON `pass` actions, not `skip`, and therefore still enforced exact `0755` before/after.
- This does not conceal a production change: the diff modifies only `sandbox_test.go`. The shared `writeFilePreserveMode` implementation still snapshots the existing mode and applies it to the synced temp before rename; neither write nor apply-patch production flow changed.
- The exact base binary independently reproduced both false failures under the same fresh Wine prefix: `expected mode 0755, got 666`. The candidate then passed both tests 100/100. That counterfactual isolates unsupported test expectation rather than a behavior change.
- PR/Issue text accurately locks the SHA/tree, Windows mapping, unchanged production scope, and separate #76 authority defect.

Independent gates with Go 1.24.5:

- Linux owning tests: count 100 PASS; race count 20 PASS. A JSON run proved both tests executed and were not skipped.
- Linux full uncached normal: 13/13 packages PASS.
- Linux full race: 13/13 packages PASS.
- Linux full vet/build, module verification, and `git diff --check`: PASS.
- Windows/amd64 with `CGO_ENABLED=0`: full vet PASS; full build PASS; sandbox test-binary compile PASS.
- Fresh Wine 9 prefix: both owning tests count 100 PASS. Full sandbox with the 13 known #76 symlink-authority tests excluded PASS.
- An unfiltered Wine run failed only the twelve invalid #76 fixtures; the thirteenth known false-green symlink test still passed without an observable link. Both mode tests passed. Thus #76 remains visibly separate and is not being used to waive #78.

Remote head/base remained exact immediately before this comment, candidate tree equals the requested-base merge-tree, and the review worktree is clean. No correctness, portability, coverage, documentation, or acceptance blocker remains for #78.

I did not modify, commit, merge, or close anything, and did not use or query GitHub Actions.

Issue #84, opened 2026-09-01
[agent/tools] 用 ToolBlockState 统一并证明 exactly-once Tool Result closure


timwhitez:
## 当前状态：顺序 terminal authority 已完成；Scheduler 集成仍开放

以 SDK `7994605` 为基线重新核对 `sdk/agent/tool_block_state.go`、Driver 与相关测试。不是“缺少 Tool Result”，也不再需要第二个 closure writer。

## 已实现

- #91–#93：duplicate-ID fail-closed 与 continuation topology cleanup；#96/#97/#101/#103：closure 与执行知识 characterization/初期观察。
- #106：运行时提交点收敛到 query-local `commitToolHistory`。#114/#115：共享 `toolResultProjection`，联合测试 handler outcome、history、ToolResultEvent、Accounting、Artifact 投影及投影失败。
- #118：`toolBlockState.acceptResults` 按 block ordinal 原子接受整批 terminal；Driver 通过唯一 writer 写 history，`takePublication` 一次领取已接受记录。重复/乱序/非法 transition 返回结构错误；拒绝路径不改已有 terminal，剩余调用保守闭合。state 由顺序 Driver 持有，不是并发 worker API。
- #120/#122/#123 与 Goode #198/#199/#202：checked active history/manual compaction/source-checked publication，合法 System-only 更新保留；已提交后 live publication 拒绝保持 indeterminate/quarantine。legacy checkpoint-only/外部 writer 的限制不因此消失。
- #83 的 Frame invocation/dispatch authority 与 #125 FrameID/InvokeAttempt 已实现；#137/#138 及 Goode #217/#218 增加显式完整 wrapper/三 builtin leaf 配置绑定。未知模型/transport/closure 仍非完整 Run 冻结，不能再把所有具体模型绑定列为未实现。

证据入口：`tool_terminal_authority_test.go`、`tool_block_admission_test.go`、`tool_block_state_test.go`、`agent_tool_block_closure_characterization_test.go`、`tool_result_projection_test.go`、`tool_outcome_projection_test.go`，以及 checked history/compaction tests。已合并 PR 的独立 Review 与 mutation 记录保留在 PR 中，不作为新的永久工作流要求。

## 语义与兼容性

[...]

PR #91, opened 2026-09-01
Reject duplicate tool call IDs before execution

timwhitez:
First safety slice of #84; this PR does not close the multi-phase Issue.

## Behavior

- synthesize missing IDs collision-safely in buffered and streaming paths;
- after recording billed Usage, reject a provider Tool Block containing duplicate non-empty IDs;
- do not append the malformed Assistant Tool Block;
- do not emit tool lifecycle events or invoke any Handler;
- do not echo the provider-supplied duplicate ID in diagnostics;
- allow a later Query to recover from clean history.

## Non-goals

No ToolBlockState, Scheduler, concurrency, cancellation semantics, Provider serializer changes, Prompt changes, or persistence changes.

## Local validation at ce8e6320adb6c67c021d7e4bcfa79f1de77c43d4 on Go 1.22.12

- go test ./... -count=1
- go vet ./...
- go test -race ./sdk/agent -count=1
- focused duplicate/missing/stream collision tests x20
- git diff --check

All passed. The earlier 52fd790 exact head received REQUEST_CHANGES; its streaming blocker is fixed and a fresh exact-head review is required before merge.

### DIFF
--- a/agent_docs/architecture.md
+++ b/agent_docs/architecture.md
@@ -74,6 +74,9 @@ Entry point: `QueryStreamWithSteering` (`sdk/agent/agent.go:197`).
      thinking blocks retain their opaque signature alongside the exact thinking
      text so a following tool-result request can serialize the assistant turn
      without dropping provider-required state.
+   - A provider response containing duplicate non-empty tool-call IDs is
+     rejected before the assistant block enters history or any handler runs;
+     already-reported usage remains accounted.
    - References: `sdk/agent/agent.go:268`
 
 7. **Continuation gate**
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -1708,19 +1708,16 @@ func duplicateToolCallIDPositions(toolCalls []llm.ToolCall) (first, second int, 
 
 func (a *toolCallAccumulator) finalize() []llm.ToolCall {
 	out := []llm.ToolCall{}
-	for i, it := range a.items {
+	for _, it := range a.items {
 		name := strings.TrimSpace(it.name.String())
 		args := strings.TrimSpace(it.args.String())
 		if name == "" {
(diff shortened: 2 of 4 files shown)


timwhitez:
REVIEW_FIX applied at exact head ce8e6320adb6c67c021d7e4bcfa79f1de77c43d4.

Resolved the 52fd790 streaming compatibility blocker by making toolCallAccumulator.finalize reuse the same collision-aware synthetic-ID allocator as buffered completions. Added a durable streaming regression where one call lacks an ID and a sibling explicitly uses call_0; both now execute with unique IDs and the explicit ID is preserved.

Fresh Go 1.22.12 exact-head gates: full tests PASS, vet PASS, sdk/agent race PASS, focused duplicate/missing/stream collision tests x20 PASS, diff check PASS. Fresh exact-head Review is in progress.

timwhitez:
INDEPENDENT_REVIEW: APPROVE

Exact base: f491e656842ebb88d83ae69a1cebb235401256e6
Exact head: ce8e6320adb6c67c021d7e4bcfa79f1de77c43d4

No blockers. The prior streaming collision finding is fixed: missing ID plus explicit call_0 now produces unique IDs, preserves the explicit ID, and executes both handlers once. Duplicate non-empty IDs fail closed after Usage/Accounting but before history, tool events, or handlers; diagnostics omit raw IDs and the next Query recovers. Focused and race matrices passed x50 on Go 1.22.12.

timwhitez:
Post-merge review correction: after merge, the reviewer found a continuation orphan path not covered by ce8e632. The defect is tracked under still-open #84 and repaired by follow-up PR #92; PR #92 removes prior partial ToolCalls before the duplicate-block terminal error and adds a persistent next-Query regression.

PR #96, opened 2026-09-01
Characterize mixed tool block closure

timwhitez:
Characterization Tests phase for #84. This PR adds one test file and changes no production code.

The mixed block freezes the current sequential closure contract across:

`unknown → panic → ordinary handler error → done/TaskCompleteError → unstarted tail`

It proves:

- all five accepted Calls keep model ID/order;
- invalid fallback, panic, ordinary error, and done handlers each start once;
- the tail handler never starts after done;
- five contiguous Provider-visible Tool Results exist exactly once and in model order;
- current result text and IsError values remain exact;
- `repairToolCallPairsDetailed` reports no repair;
- final TaskComplete content remains unchanged.

Non-goals: no ToolBlockState, shadow observer, production assertion, event/accounting/artifact change, indeterminate semantics, Scheduler, concurrency, Prompt, Provider payload, or persistence change. #84 remains open after this phase.

Local validation on Go 1.22.12:

- focused characterization test x100
- `go test ./... -count=1`
- `go vet ./...`
- `go test -race ./sdk/agent -count=1`
- `git diff --check`

All passed. Benchmark does not apply to a test-only behavior-equivalence slice.

PR #97, opened 2026-09-02
Observe sequential tool block closure state

timwhitez:
Observe-only Shadow phase for #84, built on the merged characterization PR #96.

## Scope

- adds a package-private, query-local `toolBlockState` with accepted/running/terminal phases;
- identifies Calls only by block ordinal; IDs are validated at admission but are neither retained in diagnostics nor used across blocks;
- mirrors current sequential closure paths: pre-start cancellation, loop guard, evidence suppression, Handler execution, TaskComplete, post-Handler cancellation, steering, and unstarted tail closure;
- validates at the common block exit and again through a query-level defer for future forgotten early returns;
- reports mismatch only through existing `Warningf` and does not add/change typed Events.

## Compatibility and security

The shadow does not construct or mutate history, Tool Results, events, accounting, artifacts, Provider requests, Prompt text, scheduling, or execution order. It stores no arguments, results, Prompt/source text, history index, artifact payload, secret, or cross-block/global ID. Existing mixed-block result text/order/IsError and outbound-repair=0 characterization remain exact.

## Non-goals

No enforcement, unified closure writer, `indeterminate` semantics, durable ledger, FrameID, Scheduler, concurrency, EventEnvelope projection, Prompt/Provider/persistence change, or Goode consumer. #84 remains open.

## Local validation at e8c7286

Go 1.22.12:

- focused state + mixed/cancellation/steering/loop-guard/evidence/continuation tests x50
- legacy/enveloped payload parity x50
- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`

All passed. Benchmark does not apply to an observe-only constant-work mirror.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,6 +37,10 @@ and progressive disclosure through `agent_docs/`.
   kind/origin, and observational timestamp. Legacy `QueryStream` uses the same
   direct delivery/backpressure path without metadata; sequence gaps expose
   dropped events (`sdk/agent/event_envelope.go`, `sdk/agent/agent.go`).
+- The sequential Tool Loop mirrors accepted/running/terminal transitions into
+  a query-local, observe-only `toolBlockState`. Mismatches use `Warningf`; the
+  shadow does not construct history, events, Tool Results, or Provider payloads
+  (`sdk/agent/tool_block_state.go`, `sdk/agent/agent.go`).
 - Auto-continue on `max_tokens` (including partial tool-call merge) now emits metadata-only `AutoContinueEvent` instead of text markers (`sdk/agent/agent.go:259`, `sdk/agent/agent.go:275`, `sdk/agent/agent.go:298`, `sdk/agent/events.go:113`)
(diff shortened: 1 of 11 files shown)


timwhitez:
Resolved the exact-head review test gaps at e8c72860b396b0608347d37c1fff559039b6f48d. Added deterministic coverage for root cancellation after block acceptance but before the first Handler, and cancellation triggered during TaskComplete classification before tail closure. Deleting either hook now fails 5/5. Added a two-block same-query regression proving the same tool_call_id may be reused across blocks while each ordinal-local block closes without outbound repair. Focused x50, full go test, go vet, agent race, and diff check pass on Go 1.22.12.

timwhitez:
Independent exact-head review approved e8c72860b396b0608347d37c1fff559039b6f48d against base 92a63bb032503c1834c5a0a1e1232ea827702edb. New counterexamples x100 passed; deleting either cancellation hook and replacing ordinal-local identity with cross-block global ID lookup each failed precisely. Focused/parity x50, full go test, go vet, agent race, diff, privacy, compatibility, backpressure, and accounting reviews passed. INDEPENDENT_REVIEW: APPROVE.

PR #99, opened 2026-09-02
Characterize repeated-signature intervention transcript

timwhitez:
Characterization Tests phase for #87. Relative to current main this PR changes only `agent_retry_loop_guard_test.go`; production code is untouched.

## Frozen current behavior

- four Provider requests preserve the exact system/user/assistant/tool transcript around a third-call repeated-signature intervention, including Assistant name/content and call adjacency;
- Tool definitions and ToolChoice remain stable across the intervention;
- the blocked Tool Result history text, event text, IsError, exact metadata map, and `Hidden → Warn → StepStart → ToolCall → ToolResult → Accounting → StepComplete` order are exact;
- the internal reminder appears only after the complete Tool Result block under `sdk_internal_loop_guard`;
- a non-retryable Provider failure after the reminder remains one terminal `invalid_request/400` with no Final;
- repeated-signature protection is disabled by default, so the third Handler executes and no reminder/warning is emitted.

## Non-goals

No Detector/Decision/Application production type, Intervention event schema, Prompt/Provider payload change, ToolBlock writer change, Frame lineage, evidence/idle/continuation/RequireDone migration, Thinking-only detector, persistence change, or Goode pin update. #87 remains open.

## Local validation at d60e530

Go 1.22.12:

- all `TestRepeatToolSignatureGuard*` tests x100
- `go test ./...`
- `go vet ./...`
- `go test -race ./sdk/agent`
- `git diff --check`
- benchmark x10: median `46.31 ns/op`, `32 B/op`, reported `0 allocs/op`

Benchmark command:

`go test ./sdk/agent -run ^$ -bench ^BenchmarkRepeatedToolSignatureGuardObserve$ -benchmem -count=10`

No production performance claim is made; this is the baseline for later shadow/refactor comparison.

### DIFF
--- a/sdk/agent/agent_retry_loop_guard_test.go
+++ b/sdk/agent/agent_retry_loop_guard_test.go
@@ -535,7 +535,8 @@ func TestRepeatToolSignatureGuardProviderHistoryCharacterization(t *testing.T) {
 		if !ok || result.ToolCallID != "call-3" {
 			continue
 		}
-		if !result.IsError || result.Result != "[ERROR] Tool call skipped by loop guard - Repeated identical tool call blocked before execution." || result.Metadata["loop_guard_suppressed"] != true {
+		wantMetadata := map[string]any{"loop_guard_suppressed": true}
+		if !result.IsError || result.Result != "[ERROR] Tool call skipped by loop guard - Repeated identical tool call blocked before execution." || !reflect.DeepEqual(result.Metadata, wantMetadata) {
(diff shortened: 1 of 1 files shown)


timwhitez:
Addressed the independent review blockers in d60e530: the transcript projection now freezes every Assistant message name/content separately from its Tool Calls, and the blocked ToolResult metadata must equal the complete expected map. Focused x100, full test, vet, agent race, diff-check, and the x10 benchmark pass locally. Requesting exact-head re-review.

PR #100, opened 2026-09-02
Characterize provider admission identity boundaries

timwhitez:
Characterization phase for #83. Relative to current main this PR changes only `agent_request_ownership_test.go`; production code is untouched.

## Frozen current behavior

- one framework retry loop captures the same outer `llm.ChatModel` interface across attempts;
- the next independent admission captures the newly installed outer model;
- a dynamic outer wrapper may still change its concrete target between attempts, explicitly recording the remaining snapshot gap;
- every attempt receives an independently cloned, provider-neutral request that remains exactly equivalent across messages, nil/empty schemas, concrete JSON number types, ToolChoice, Temperature, thinking flag, and all mutable Responses pointer/slice/map options;
- the first Provider attempt mutates nested request and Responses state, proving the next attempt does not inherit those mutations;
- `BenchmarkProviderAdmissionSnapshot` records the current clone + no-op Provider admission baseline.

## Non-goals

No production ExecutionFrame, FrameID/lineage, fingerprint, observer, public API, concrete-model snapshot, tool resolver refactor, Prompt/Provider payload, Tool execution, persistence, compaction behavior, or Goode pin update. #83 remains open.

## Local validation at 631fbe9

Go 1.22.12:

- focused admission/ownership tests x100
- `go test ./...`
- `go vet ./...`
- `go test -race ./sdk/agent`
- `git diff --check`
- outer-model re-read mutation killed 20/20
- Store + Reasoning shallow-clone mutations killed 20/20
- benchmark x10: median `4694 ns/op`, `2240 B/op`, `50 allocs/op`

Benchmark command:

`go test ./sdk/agent -run ^$ -bench ^BenchmarkProviderAdmissionSnapshot$ -benchmem -count=10`

No performance claim is made; this is the baseline for later shadow/refactor comparison.

### DIFF
--- a/sdk/agent/agent_request_ownership_test.go
+++ b/sdk/agent/agent_request_ownership_test.go
@@ -119,6 +119,7 @@ func (m *providerAdmissionModel) Invoke(_ context.Context, request llm.InvokeReq
 	if call == 1 && m.retryFirst {
 		request.Messages[0].Content.Blocks[0].Text = "mutated"
 		request.Tools[1].Parameters["limit"] = int64(99)
+		request.Tools[1].Parameters["items"].([]any)[0] = "mutated"
 		*request.Temperature = 9
 		*request.Responses.UseResponseItems = false
 		*request.Responses.UseInstructions = true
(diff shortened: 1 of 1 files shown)


timwhitez:
Addressed the independent review blocker in f2cc0b0. The first attempt now mutates Temperature plus every mutable Responses pointer/slice/map family, including Store, Reasoning, UseInstructions, Include, Text/Format, and OutputSchema. The reviewer Store+Reasoning shallow-clone mutations now fail 20/20. Focused x100, full test, vet, agent race, diff-check, and benchmark x10 pass. Requesting exact-head re-review.

timwhitez:
Addressed the second independent review blocker in 631fbe9. Assertions now compare every recorded attempt to a newly constructed golden that cannot share state with the request under test. Both Tool Parameters and Responses schema contain non-empty slices mutated in place. The reviewer Reasoning, Include, and generic JSON-slice shallow-copy mutations now fail; focused x100, full test, vet, agent race, diff-check, and benchmark x10 pass. Requesting exact-head re-review.

PR #101, opened 2026-09-02
Characterize running-tool cancellation outcome

timwhitez:
Additional Characterization phase for #84. Relative to current main this PR changes only `agent_cancel_boundary_test.go`; production code is untouched.

## Frozen current behavior

- a Handler may perform its side effect and then cancel the root context, whether it returns success or an ordinary error;
- the already-running Call is currently closed with the exact turn-continuation cancellation result, so this legacy text must not be interpreted as proof that the side effect did not happen;
- an unstarted sibling is closed separately with the exact `before this call ran` result and its Handler never starts;
- the running Call emits exact StepStart → ToolCall → ToolResult → Accounting → StepComplete → canceled ordering;
- the unstarted sibling currently emits no Tool/Step/Accounting lifecycle event;
- the two Provider-history Tool Results remain contiguous and require zero outbound repair;
- root cancellation prevents same-Turn Provider re-admission, while the next Query can send the already-closed block without a repair warning;
- `BenchmarkToolBlockShadowLifecycle` records an 8-call accepted/running/terminal shadow baseline.

## Non-goals

No production outcome/state field, indeterminate shadow, single writer, History/Prompt/Provider payload/event change, FrameID, Scheduler, parallel/batch execution, Artifact or persistence change. #84 remains open.

## Local validation at 795db9c

Go 1.22.12:

- focused cancellation/closure tests x100
- `go test ./...`
- `go vet ./...`
- `go test -race ./sdk/agent`
- `git diff --check`
- benchmark x10: median about `1530 ns/op`, `368 B/op`, `2 allocs/op`

Benchmark command:

`go test ./sdk/agent -run ^$ -bench ^BenchmarkToolBlockShadowLifecycle$ -benchmem -count=10`

No performance claim is made; this is the baseline for the later indeterminate shadow and writer migration.

### DIFF
--- a/sdk/agent/agent_cancel_boundary_test.go
+++ b/sdk/agent/agent_cancel_boundary_test.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"errors"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
+	sdkaccounting "github.com/timwhitez/agent-sdk-golang/sdk/accounting"
 	"github.com/timwhitez/agent-sdk-golang/sdk/llm"
 	"github.com/timwhitez/agent-sdk-golang/sdk/tools"
 )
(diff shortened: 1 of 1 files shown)


timwhitez:
Addressed the independent review blocker in 795db9c. The second Provider request must now contain the exact previously closed running+tail block, not merely any internally contiguous history. I also made ToolResult/Accounting adjacency explicit using raw event indexes. Focused x100, full test, vet, agent race, diff-check, and benchmark x10 pass. Requesting exact-head re-review.

PR #103, opened 2026-09-02
agent: observe indeterminate running-tool cancellation

timwhitez:
Progresses #84.

## Scope

- extend the existing query-local `toolBlockState` with private execution knowledge: unknown, not-started, attempt-started, outcome-observed, and indeterminate;
- mark a returned running attempt indeterminate only when root cancellation causes the legacy handler outcome to be overwritten;
- keep unstarted accepted siblings explicitly not-started;
- validate that terminal running calls have observed or indeterminate knowledge;
- expose the final private shadow state only to a package-private test observer so runtime wiring is mutation-tested.

`outcome_observed` means only that the SDK retained a return/error/panic outcome. It does not claim an external side effect was applied or not applied.

## Non-goals

No Tool Result/history/event/accounting/provider payload changes; no public API or persistence; no single writer; no scheduler/concurrency; no Frame/EventEnvelope projection. A context-ignoring handler that never returns remains outside this slice.

## Local verification

Exact base: `bee696f7113e0f58102985dfabe17bf9cb08c587`
Exact head: `72d46333b9c44c3635b34e08209fa96e944ccc86`

- focused state/cancellation tests x100
- focused state/cancellation race x20
- `go test ./... -count=1`
- `go vet ./...`
- `BenchmarkToolBlockShadowLifecycle` x10: 2 allocs/op, 432 B/op (baseline 368 B/op; +64 B for eight ordinal knowledge fields)
- `git diff --check`

Independent exact-head review is required before merge.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -169,6 +169,7 @@ type Agent struct {
 
 	compactionAdmissionObserved func()
 	compactionShadowObserved    func()
+	toolBlockStateObserved      func(*toolBlockState)
 
 	tools             []tools.Tool
 	toolMap           map[string]tools.Tool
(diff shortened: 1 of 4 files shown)


timwhitez:
Review remediation at exact head `72d46333b9c44c3635b34e08209fa96e944ccc86`:

- added a direct terminal/unknown invariant oracle;
- added real QueryStream observation for ordinary handler success, error, and panic, each requiring `outcome_observed`;
- reran focused x100, full, vet, agent race, focused race x20, and diff-check.

Independent exact-head re-review requested.

PR #104, opened 2026-09-02
agent: shadow repeated-signature interventions

timwhitez:
Progresses #87.

## Scope

- add private typed Detection and Decision shadows for the existing repeated-signature guard only;
- model proceed/suppress-tool, reminder queueing, and guard downgrade from captured scalar facts;
- call stateful signature observation once and cache the existing recycled-result lookup;
- keep legacy `blocked`, strike/reset, history, events, Tool Results, and continuation as the sole behavior authority;
- emit only closed kind/action labels and booleans on mismatch;
- use package-private test seams to cover exhausted/recycled runtime inputs and deliberately divergent shadow decisions.

## Non-goals

No Application type or writer; no Prompt/history/event/provider payload changes; no thresholds or guard behavior changes; no idle, RequireDone, evidence recovery, retry, steering/cancel, compaction, public API, persistence, Frame, or EventEnvelope integration.

## Local verification

Exact base: `d2616fceb7187ac226962a9c9b1379253a7a584f`
Exact head: `82cda0fa9d424896f1a0c6001a385958c86edd1d`

- new shadow/runtime focused tests x100
- existing repeat Provider transcript/event/failure/default/downgrade tests x100
- focused repeat race x20
- `go test ./... -count=1`
- `go vet ./...`
- `BenchmarkShadowRepeatedSignatureIntervention` x10: median 4.588 ns/op, 0 B/op, 0 allocs/op
- `git diff --check`

Independent exact-head review is required before merge.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -171,6 +171,8 @@ type Agent struct {
 	compactionShadowObserved    func()
 	toolBlockStateObserved      func(*toolBlockState)
 
+	repeatInterventionShadowObserved func(interventionDecision, interventionDecision)
+
 	tools             []tools.Tool
 	toolMap           map[string]tools.Tool
 	toolMapNormalized map[string]tools.Tool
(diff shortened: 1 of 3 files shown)


timwhitez:
Review remediation at exact head `82cda0fa9d424896f1a0c6001a385958c86edd1d`:

- added real runtime decision oracles across strike downgrade, exhausted-normal proceed, and exhausted-recycled suppress boundaries;
- assert the recycled-result lookup occurs exactly once at that boundary;
- added a package-private evaluator seam and deliberately divergent shadow, proving the full legacy Provider transcript, event order, Tool Result, and execution count remain unchanged;
- reran focused x100, full, vet, agent race, focused race x20, benchmark, and diff-check.

Independent exact-head re-review requested.

PR #105, opened 2026-09-02
agent: observe conservative tool-call plans

timwhitez:
Advances #85 without closing the multi-phase issue.

## Scope

- add a package-private per-call shadow planning observation after legacy tool resolution and argument normalization
- classify every reached call as `Exclusive` with its original ordinal
- compare the shadow result with the still-authoritative sequential loop and emit only an allowlisted mismatch warning
- cover exact/alias/fallback resolution, invalid arguments, done/cancel/duplicate-ID boundaries, repeat/evidence suppression, non-authoritative mismatches, and secret-negative diagnostics

## Safety / non-goals

- no production concurrency, waves, batching, traits, conflict keys, or public API
- no tool-name/schema/prompt semantic guessing; observation contains no names, IDs, arguments, schemas, handlers, deps, or context
- no changes to Provider payloads, Tool Results/history/events/accounting, retry, steering, cancellation, compaction, or Goode batch behavior
- the legacy sequential loop remains the sole execution authority
- the first normalization is observational only; `Tool.Execute` still performs the existing second normalization/possible schema-key repair

## Local verification

- `gofmt -w sdk/agent/agent.go sdk/agent/tool_plan_shadow.go sdk/agent/tool_plan_shadow_test.go`
- focused runtime/closure/cancel/guard suite `-count=100`
- `go test ./... -count=1`
- `go vet ./...`
- `BenchmarkShadowToolCallPlan -benchmem -count=10`: median about 0.265 ns/op, 0 B/op, 0 allocs/op
- `git diff --check`

Exact base: `3680d808515c1c983b67a9a05ce0dad1effa7a55`
Exact head: `2f5276a14a575c2c5ac0a3ae1706eeda8ec20728`

Independent review approved this head. Eight mutation categories were rejected by tests, including the original read/batch classification survivor. Focused x100, full tests, vet, and agent race passed on this commit.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -174,6 +174,7 @@ type Agent struct {
 	repeatInterventionShadowObserved  func(interventionDecision, interventionDecision)
 	repeatInterventionShadowEvaluator func(repeatedSignatureObservation) interventionDecision
 	repeatResultRecycled              func(string) bool
+	toolPlanShadowEvaluator           func(toolPlanningObservation) toolCallPlan
 
 	tools             []tools.Tool
 	toolMap           map[string]tools.Tool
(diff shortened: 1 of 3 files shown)


timwhitez:
Review follow-up at `2f5276a14a575c2c5ac0a3ae1706eeda8ec20728` adds a runtime regression requiring both alias-resolved `read` and host `batch` to retain the default Exclusive plan with no planner mismatch warning. The earlier review found that a name-based classification mutation survived because the existing fixture injected its own shadow result.

Local validation on the updated tree passed: focused planner tests x100, `go test ./... -count=1`, `go vet ./...`, `go test -race ./sdk/agent -count=1`, gofmt, and diff check. A fresh independent review of this commit is in progress. Issue #85 remains open for the remaining Scheduler stages.

PR #106, opened 2026-09-07
refactor(agent): centralize Tool Result history closure

timwhitez:
## Scope

Advances #84; does not close the multi-phase issue.

Centralizes the nine accepted Tool Result history commits through one query-local writer, preserving append-before-shadow-terminal ordering. Reuses existing message construction and removes redundant append wrappers. Adds synthetic-ID, done-not-last, history flags/text, terminal execution knowledge, and history-only tail regression coverage.

No changes to provider-visible text/payload, execution order, event/accounting/artifact boundaries, persistence, public APIs, or concurrency. Duplicate-completion enforcement and shared terminal-outcome projections remain future slices.

## Local validation at fbea5d836d247edc1d319faae4533d2bc97e9ae2

- gofmt and git diff --check
- focused closure/cancellation/guard characterization tests x100
- go test ./... -count=1
- go vet ./...
- go test -race ./sdk/agent -count=1
- BenchmarkToolBlockShadowLifecycle: five samples, median 722.6 ns/op, 432 B/op, 2 allocs/op; this measures the existing state lifecycle, not an end-to-end speedup.

Go 1.22.12. Independent exact-head review APPROVE: all nine paths checked, full/vet/focused x20 passed and four mutations detected. Initial independent agent race failed on the existing zero-event steering fixture; base reproduces 60 failures in race x1000. Tracked and fixed separately in #107/#108. Subsequent independent same-head agent race single run and x3 passed. See the review comment for evidence; the initial failure is not hidden.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -683,6 +683,14 @@ func (a *Agent) queryStreamWithSteering(ctx context.Context, input llm.Content, 
 			emitErr(e, origin)
 		}
 		var activeToolBlock *toolBlockState
+		// All accepted Tool Result history commits pass through this writer.
+		// Event/accounting projection stays at each existing emission boundary.
+		closeToolResults := func(start int, expected toolCallPhase, reason string, messages ...llm.Message) {
+			a.appendMessages(messages)
+			for offset := range messages {
+				activeToolBlock.markTerminal(start+offset, expected, reason)
+			}
+		}
 		finishToolBlock := func() {
 			block := activeToolBlock
 			activeToolBlock = nil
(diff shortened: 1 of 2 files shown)


timwhitez:
Independent exact-head review: APPROVE for fbea5d836d247edc1d319faae4533d2bc97e9ae2.

All nine closure paths checked against base: history fields, append lock granularity, ordinal, event/accounting boundaries preserved. Independent full tests/vet/focused x20 pass. Four mutation checks (missing/duplicate history, wrong ordinal, missing terminal transition) fail as expected.

Initial agent race failed in an unrelated existing test, TestSteeringInterruptErrorIsNotRetryable:473. The zero-event streaming fixture races normal Done/close against preloaded steering. Base 577a36f reproduces 60 failures in race x1000, and the failing direct invoke path does not run this PR's closure code. Tracked separately as #107; fixing before final-main verification.

Subsequent independent exact-head agent race single run and x3 pass. These do not erase the initial failure; its cause is explicitly tracked above.

PR #108, opened 2026-09-07
test(agent): make steering retry policy check deterministic

### LINKED ISSUE #107: test(agent): remove completion race from steering retry policy test
## Confirmed main failure

At SDK main 577a36fa603fbdc647cfb65a33e4bec26efc8282, `go test -race ./sdk/agent -run '^TestSteeringInterruptErrorIsNotRetryable$' -count=1000` reproduces 60 failures in independent review: expected SteeringInterruptError, got nil. This also caused the initial full race gate for #106 to fail; that PR does not touch the failing invocation path.

## Cause

The test uses `slowStreamingModel{}` with zero events. It immediately emits Done and closes the stream, competing with preloaded steering in select. Normal provider completion can legitimately win. The elapsed-time assertion also does not prove the number of retry attempts.

## Narrow fix and acceptance

Make the retry-policy test deterministic by reusing the existing invocation-error fixture with a typed SteeringInterruptError. Assert exact invocation count and retained typed error; include retry-looking text to ensure the explicit steering guard takes precedence over generic classification. Existing streaming-interruption tests retain stream integration coverage.

No production/provider/prompt changes. Require focused repeated race, full tests/vet/agent race and independent exact-head review.

timwhitez:
Closes #107.

Reuses the existing deterministic invocation-error fixture for the steering retry policy test. Checks one provider call and preservation of the exact typed interruption, including retry-looking HTTP 429/timeout text. Removes the elapsed-time proxy and the zero-event stream-vs-steering select race. Existing stream interruption/drain tests retain integration coverage.

No production changes; one test file, net nine lines removed.

Validation: gofmt, focused race x1000, go test ./... -count=1, go vet ./..., go test -race ./sdk/agent -count=1, git diff --check (Go 1.22.12). Independent exact-head review required before merge.

### DIFF
--- a/sdk/agent/agent_steering_interrupt_test.go
+++ b/sdk/agent/agent_steering_interrupt_test.go
@@ -439,8 +439,11 @@ func TestInvokeCompletionWithSteeringNilChannel(t *testing.T) {
 // TestSteeringInterruptErrorIsNotRetryable verifies that steering
 // interrupt errors are not retried.
 func TestSteeringInterruptErrorIsNotRetryable(t *testing.T) {
+	// Retry-looking text must not override the typed steering disposition.
+	interrupt := &llm.SteeringInterruptError{Message: "HTTP 429 timeout"}
+	model := &transientInvokeModel{failFor: 5, err: interrupt}
 	ag, err := New(Config{
-		LLM:                    &slowStreamingModel{},
+		LLM:                    model,
 		InvokeRetryMaxAttempts: 5,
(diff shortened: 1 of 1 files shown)


timwhitez:
Independent exact-head APPROVE at 7cb61a9999ba7fe3883a68fe4422f43901ccaa09. Full tests, vet, agent race, focused race x1000, gofmt/diff-check pass. Removing the typed-steering early return in an isolated mutation tree makes the test fail with provider calls=5, want 1. Existing streaming integration coverage retained.

PR #109, opened 2026-09-07
feat(agent): shadow logical requests with execution frames

timwhitez:
Advances #83; does not close it.

## Scope

- Private executionFrame owns a cloned SDK logical request and exact/normalized resolver schemas, retaining outer model and handler handles.
- Shared existing resolver implementation; preserve exact/alias/collision/hidden and registered/internal invalid fallback compatibility.
- Observe advertised definitions and dispatch metadata through fixed structural Warningf messages only. No request/schema/raw tool names/results are logged.
- Legacy invoke/dispatch remains authoritative. Continuation dispatch uses the finalizing request snapshot, not a claim that merged arguments have only one request source.
- Warningf may synchronously cancel: actual provider/tool admission gates remain after diagnostics. Tests cover unstarted cancellation closure.
- Correct outdated AGENTS/CLAUDE claim that New injects invalid into its registry.

No FrameID, fingerprint, public API, second event sequence, concrete dynamic-model snapshot, handler closure-identity proof, serializer/Prompt/persistence change or production parallelism. Model/handler closure state and dependency environment remain mutable.

## Tests and cost

Ownership, alias/hidden/fallback, safe diagnostics, clone-failure compatibility, legacy dispatch despite drift, cancellation, continuation finalizing request and repair=0 regressions; focused x100 passes.

Full test/vet/agent race passed at b37d495. Independent review found two surviving runtime-wiring mutations (deleted advertisement observer / empty request capture). Fixed in 1a9d016446a702f3d695397a5b2ae165ac36b9cd with a real QueryStream advertisement/request/legacy-result/cancellation fixture. Final-head author and independent full/vet/agent race/focused x100 passed; independent re-review APPROVE confirmed both mutations are detected. Merged as e5ce39dc1e419b680e4199104e521ad71c5e519a.

Five local microbenchmark samples (Go1.22.12, i9-13900H): existing admission clone median 5871 ns/op, 2240 B/op, 50 allocs/op; new frame constructor median 9964 ns/op, 4688 B/op, 66 allocs/op. These are different operations, not an end-to-end speedup comparison: shadow adds allocation/clone overhead per logical request. Reassess representative registry/history scaling before enabling authority or claiming performance improvements.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -99,7 +99,16 @@ agent_docs/          Detailed architecture/provider/tool/build/test documentatio
 
 ### Configuration
 - Configure behavior via `agent.Config` (LLM, tools, tool choice, compaction, history, dependency context) (`sdk/agent/agent.go:20`)
-- `Agent.New` injects hidden `invalid` fallback tool when absent (`sdk/agent/agent.go:68`, `sdk/tools/invalid.go:15`)
+- Unknown tool dispatch uses a registered `invalid` tool when present, otherwise
+  constructs the internal fallback at dispatch; `Agent.New` does not add it to
+  the advertised tool list (`sdk/agent/agent.go`).
+- A private `executionFrame` shadows each logical request with owned request
+  and resolver schemas plus captured model/handler handles. Legacy invocation
+  and dispatch remain authoritative. Continued arguments can have multiple
+  request sources; the dispatch shadow belongs to the finalizing request.
+  Diagnostics use structural `Warningf` messages only, not a second event path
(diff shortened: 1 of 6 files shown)


timwhitez:
Independent exact-head APPROVE at 1a9d016446a702f3d695397a5b2ae165ac36b9cd after re-review. Full tests, vet, agent race, focused x100 and formatting/diff checks passed. The two initial surviving mutations (deleted runtime advertisement observer and empty request capture) are now detected by the real QueryStream test. Resolution observer and tool-cancellation gate mutations were also detected. No unresolved findings. Benchmarks are harness-only microbenchmarks, not real model/provider performance or failure scoring.

PR #128, opened 2026-09-08
feat(llm): add pure cache target view and snapshot-bound validation

timwhitez:
## Scope

Related to #89 (multi-phase, remains open). Builds on merged #127 characterization instead of repeating it.

New opt-in pure CacheTargetView owns a CloneInvokeRequest snapshot, exposes content-free logical source coordinates, and validates a plan against the retained snapshot and current request. No production provider/Agent caller is added. No wire, prompt, history, model selection, scheduling or persistence change.

Binding uses conservative exact in-memory Go-value equality, not new content fingerprints or commit gates. Existing experimental fingerprint fields remain source-compatible but are explicitly unsupported by this checker. The caller must retain the original view; rebuilding a view after an edit cannot prove an old plan's provenance. This is not a session revision, concrete-model snapshot or TOCTOU-safe send API.

## Contract

- Ordinals: nonempty Text, addressable visible explicit blocks, assistant ToolCalls. Opaque/thinking/redacted/empty/unknown blocks are not targets. Original source indexes are retained, not confused with Provider wire offsets.
- Tool messages expose one outer result boundary, not nested content blocks. Message boundaries alias the final logical target for duplicate detection; adapters must still reject unprovable wire mappings, including hidden trailing blocks.
- Validates schema, current snapshot equality, kind/indices, policy/TTL syntax and duplicate/conflicting boundary aliases. Fixed typed reason+ordinal errors do not wrap content-bearing clone errors.
- Nil plan needs no validation. Nil/zero view with a plan fails. Returned metadata is owned; immutable snapshot supports concurrent reads with caller-owned input graphs.

## Validation

Author exact df7c7c7577f32b5f563dc6fa5c7c47bf56fbfccf: focused100, full/vet/LLM+provider+Agent race passed. Golden logical coordinates, stale text/options/schema/pointer/reorder/compaction, nil-empty identity, clone failure, privacy, unchanged inputs, duplicate aliases, concurrent reads and runnable example included. Existing HTTP wire goldens remain authoritative for inert plan compatibility. Independent exact-head review in progress before merge.

Benchmark (three samples, median) measures only snapshot construction / one last-boundary validation for 1, 32, 256 one-KiB messages: snapshot432.9/11988/100428ns,512/16880/138992B,5/10/13alloc; validation587.1/5721/45753ns,192B/2alloc. Strings remain shared immutable Go values. No speedup, provider/model performance or worst-case input claim. Logs /tmp/sdk-cache-view-exact-*.log.

[...]

PR #129, opened 2026-09-08
feat(llm): resolve cache intent against concrete client capabilities

timwhitez:
## Scope

Related to #89; multi-phase Issue remains open. Reuses the #128 retained CacheTargetView and existing capability interface. Adds opt-in Decide: fixed typed per-directive decisions and an owned filtered plan. It queries the supplied actual ChatModel's optional capability method once, never Provider()/Model() names or network methods.

Required directives reserve capacity first. Unsupported target/TTL or required overflow returns a fixed validation error and no partial decision. Remaining best-effort directives are deterministically retained in original order or marked skipped. Invalid/stale plans fail for either policy. Zero MaxBreakpoints means no explicit directives; default TTL uses the client default, explicit TTLs need a reported entry. Negative capacity/unknown reported TTL fails safely without returning arbitrary client strings.

Anthropic, OpenAI Chat and Responses now report the truth: optional cache-usage normalization exists, explicit CachePlan mapping does not. Legacy Message.Cache/MaxCachedToolDefinitions/implicit caching are not advertised as new explicit controls.

No provider invokes this helper automatically. No wire, prompt, persistence, Frame/event sequence, scheduling, hash gate or existing public struct layout change. Accepted means capability eligibility only, not mapped/sent/cached. The original view still must be bound to the owned request at actual sending; this helper is not a send authorization or concrete-model snapshot.

## Validation

Focused100 passed. Tests cover all target flags, TTL/default policy, required priority, overflow/no partial result, unknown clients, stale/no callback, nil/typed-nil model, conservative builtins, content-free outputs, clone isolation, callback input mutation and read-only concurrency. Existing nil/empty/inert-plan HTTP wire goldens unchanged.

Author exact head3d5144da8211dbc3fd3c2ffabbfaf944abd973a9: full/vet/LLM-provider-Agent race passed locally. Benchmark three-sample median730.0ns/op,448B/op,6allocs/op for one directive/one one-KiB message using a retained view and fake capability client (including its atomic query counter, excluding initial view creation); not provider/model latency or a speedup claim. Logs /tmp/sdk-cache-decision-{full,vet,race,bench}.log. Independent exact-head review required before merge; no live provider/cache-hit claim.

### DIFF
--- a/sdk/llm/anthropic/client.go
+++ b/sdk/llm/anthropic/client.go
@@ -73,6 +73,12 @@ func (c *Client) warnf(format string, args ...any) {
 
 func (c *Client) Provider() string { return "anthropic" }
 
+// PromptCacheCapabilities reports explicit CachePlan mapping, not legacy
+// Message.Cache/MaxCachedToolDefinitions. Those legacy controls remain separate.
+func (c *Client) PromptCacheCapabilities() llm.PromptCacheCapabilities {
+	return llm.PromptCacheCapabilities{UsageTelemetry: true}
+}
+
 func (c *Client) Model() string { return c.ModelName }
 
 func (c *Client) Invoke(ctx context.Context, req llm.InvokeRequest) (*llm.Completion, error) {
--- a/sdk/llm/cache_decision.go
+++ b/sdk/llm/cache_decision.go
@@ -1,0 +1,122 @@
+package llm
+
+import "reflect"
+
+// CacheDirectiveDecision is bounded, content-free metadata for one original
+// directive. Accepted means eligible under reported capabilities, not mapped,
+// sent, or cached. Reason is accepted, unsupported_target, unsupported_ttl, or
+// breakpoint_limit. No content, names, IDs, hashes or Provider strings are kept.
+type CacheDirectiveDecision struct {
+	DirectiveIndex int
+	Accepted       bool
+	Reason         string
+}
+
+// CachePlanDecision owns a filtered plan and decisions in original order.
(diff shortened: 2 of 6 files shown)


timwhitez:
Independent subagent pin194_review APPROVE exact3d5144da8211dbc3fd3c2ffabbfaf944abd973a9. Independent full/vet/build/LLM-provider-Agent race/focused100 passed. Five mutations detected missing required reservation, skipped TTL validation, shared plan/capability aliases, and partial result on required failure; restored clean. Goode reviewer goode121_review independently passes full/vet/five-package race/strict actual-source build against Goode3d28165, plus 3builtins x 3wrapper runtime capability assertions under race x10: no accidental interface promotion or existing Decide consumer. No author/main edits by reviewers. Approval is only pure capability decision and existing-host compatibility, not wire mapping/cache hits/send atomicity. Logs /tmp/review129-*.log and /tmp/review-goode-cachedecisions-*.log.

timwhitez:
Merged-main evidence: SDK3150d46d3026cc64e46be198649b745cb7cf5696 full/vet/LLM-provider-Agent race passed (agent40.210s); Goode3d28165 full/vet/six-package race/strict build passed against that SDK. Version stamps actual SDK3150d46 without pin update. Both main trees clean, no open PRs; #89 updated to next real before-network admission/binding and existing diagnostics reuse, not another pure helper. Logs /tmp/sdk129-main-*.log and /tmp/goode-sdk129-main-*.log. Exact mapping/host generator/full #89 remain unfinished.

PR #130, opened 2026-09-08
feat(llm): enforce bound cache intent before provider network I/O

timwhitez:
## Scope

Related to #89; this advances actual before-network admission, not another pure helper and not completion of explicit cache mapping.

CacheTargetView.Bind creates an owned nonempty plan with a private original-view binding. Bind copies the view value into unreachable private storage: the review-discovered whole-view overwrite/rebinding bypass is fixed and has external-package real-client regressions. CloneCachePlan shares only that immutable private binding; JSON cannot restore it.

All six built-in Invoke/InvokeStream entrypoints call shared AdmitCachePlan before goroutines/network. Nonempty requests use existing CloneInvokeRequest, validate the original binding, and reuse Decide. Required unsupported/stale/unbound intent fails before HTTP; best-effort unsupported intent is skipped while legacy flags/wire remain. No accepted unmapped directive is silently sent. Fixed diagnostics are capped at32 plus one summary; warning sinks receive them and buffered successes also carry Completion.Diagnostics.

Review caught another real regression in the first candidate: adding Responses.WarningSinkSetter caused concurrent shared child Agent.New to mutate one client's sink, including nil-plan flows. Final c4ed532 removes that setter entirely. Responses.Warningf is construction-time configuration (otherwise existing logger behavior), not automatically overwritten by child Agents. SDK concurrent child creation+real requests proves configured sink80/child sink0; Goode's original reproduction also passes after the fix. Buffered metadata and warning output are two observable channels, not two distinct decisions; per-child/request-local sink adoption remains future host work.

## Intentional compatibility changes

- Previously inert nonempty plans are now enforced. Construct them with view.Bind; handmade/restored unbound plans reject. Nil/empty plan legacy behavior remains, including wire bytes.
- CachePlan private binding and ResponsesClient.Warningf change public struct layouts; external unkeyed literals need migration. Existing keyed literals, Message/Completion layouts and session JSON remain compatible.
- No Prompt rewrite, explicit cache-control mapping, model snapshot, Scheduler/event sequence, persistent authority or hash gate is introduced.

## Validation

Focused100 passed. Real 3provider x buffered/streaming tests prove zero HTTP for required/unbound/stale/clone-stale/stale-best-effort/uncloneable/schema/cancel and whole-view overwrite. Tests also cover private sourceview clearing, best-effort 200/401 wire parity+diagnostics, callback-time caller mutation isolation, partial stream cancellation/body closure, JSON binding loss and diagnostic cap/privacy.

Legacy golden variants now cover nil/empty/bound best-effort parity; formerly ignored malformed/required cases move to explicit before-network rejection tests. Existing serializers are otherwise unchanged. Author final c4ed532 full/vet/LLM-provider-Agent race/focused100 passed. Original binding overwrite and new shared-sink race findings are fixed, not waived. Independent exact-head re-review required after both fixes.

Final test-only follow-up654b56f isolates Agent.New's temporary dump directory. Repeated race testing identified unrelated shared-temp index cleanup warnings contaminating the child sink count (configured80/child6 with only fixed dump-index warning formats, no data race). The fixture still asserts configured80/child0; it does not weaken the cache routing invariant. Isolated combined race100 passed; final full/vet/race and independent exact-head review refreshed before merge.

Non-goals: exact Anthropic mapping, actual cache hits, dynamic model/config snapshot, host plan generation and completed #89. Current builtin capabilities still advertise no explicit mapping.

Final-candidate admission benchmark (3-sample median):1266ns/op,880B/op,12allocs/op for one one-KiB message/one bound best-effort directive with prebuilt view, nil warning callback and no network. Includes owned clone, validation, decision and bounded diagnostic construction; excludes initial binding/Provider latency. Not a speedup/cache-hit claim.

### DIFF
--- a/agent_docs/providers.md
+++ b/agent_docs/providers.md
@@ -6,13 +6,17 @@ stream normalization, and response metadata behavior.
 ## Shared LLM Contracts
 - `InvokeRequest.CachePlan` is experimental in-memory intent, owned by
   `CloneInvokeRequest` and the private execution frame, and excluded from JSON.
-  Built-in serializers still ignore it: fingerprint/target validation,
-  required/best-effort policy and capability-gated wire mapping are not enabled.
+  Bind nonempty plans with `CacheTargetView.Bind`. Built-in buffered/streaming
+  entrypoints clone and validate that original binding before network I/O:
+  stale/unbound/unsupported Required plans fail; Best-effort skips produce bounded
+  `Completion.Diagnostics` and warning-sink metadata. Explicit wire mapping is
+  still unavailable; nil/empty plans preserve the legacy path.
   Legacy `Message.Cache` behavior is unchanged; do not use the new field as a
   cache-control guarantee. No Agent Config or host plan generator is added.
-  Adding this exported field preserves keyed literals but external unkeyed
-  `InvokeRequest` literals must be updated. Message/Completion layouts and
-  session JSON are unchanged.
+  `CachePlan` now includes a private binding and `ResponsesClient` supports the
+  existing `WarningSinkSetter` pattern. External unkeyed literals need updating;
+  JSON cannot restore a valid binding. Message/Completion layouts and session
+  JSON remain unchanged. No content fingerprints or Git-hash gates are added.
 - Provider-neutral interfaces: `ChatModel`, `StreamingChatModel` (`sdk/llm/model.go:8`, `sdk/llm/model.go:17`)
 - Unified request envelope: `InvokeRequest` (messages, tools, tool choice, temperature, responses options) (`sdk/llm/model.go:82`)
 - Unified stream event union: text/thinking/tool-call deltas, usage, done, response metadata, errors (`sdk/llm/model.go:24`, `sdk/llm/model.go:28`, `sdk/llm/model.go:33`, `sdk/llm/model.go:39`, `sdk/llm/model.go:50`, `sdk/llm/model.go:55`, `sdk/llm/model.go:62`, `sdk/llm/model.go:70`)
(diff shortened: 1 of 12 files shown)


timwhitez:
Independent SDK reviewer pin194_review APPROVE final654b56f3753f61368edbef817050920b7bb0560d, superseding REQUEST_CHANGES. Full/vet/build/LLM-provider-Agent race/admission100, combined race100, sink race10 and actual Goode shared-child probe10 passed. Reintroducing old setter still causes data race and configured0/child80: fixture isolation does not weaken the regression. View-copy fix and real best-effort overwrite rejection proven. Independent Goode reviewer also APPROVE final654: runtime diff fromc4 empty; final fixture race10/build/actual binding pass, full/vet/5pkg race reuse explicitly limited to identicalc4 runtime evidence. Both actual blockers fixed, not waived. Logs /tmp/review130-isolated-*.log and /tmp/review-sdk130-final654-*.log. Constructor-scoped Responses sink is not request-local trace identity; exact cache mapping and #89 remain unfinished.

timwhitez:
Merged-main verification passed: SDKffdb2d65a146d59dd5a731f520499a707e482736 full/vet/LLM-provider-Agent race (agent39.787s); Goode3d28165c1648cf829540c9c36a368ff65cd44f59 full/vet/six-package race/strictbuild against it. --version stamps actual SDKffdb2d6, no pin-only update. Both main trees clean; no open PRs; twelve original multi-phase Issues remain. #89 now correctly moves from real admission to precise Anthropic target mapping and later host/model/diagnostic integration. Logs /tmp/sdk130-main-*.log and /tmp/goode-sdk130-main-*.log. No live-provider/cache-hit claim.

PR #134, opened 2026-09-08
fix(agent): bind provider warnings per invocation without shared mutation

timwhitez:
## Scope / SDK issue133

Stop Agent.New from calling WarningSinkSetter on shared model handles. Bind the Agent warning sink through llm.WithWarningSink at the common model invocation path; bind the Service sink at compaction's only model call. All three builtin buffered/streaming clients capture the context sink in call-local state before admission/retry/goroutines. OpenAI reuses its existing local client copy; no global lock, shared registry, new event sequence or executor is added.

Direct clients without a binding retain configured/default logging. Nil binding masks an inherited sink and restores fallback. Manual setters remain configuration-time APIs, not concurrent runtime mutation APIs. Custom providers must read llm.WarningSink(ctx,fallback); SDK no longer implicitly configures their setter. Provider request/persistence formats are unchanged.

## Paired dependency

Goode Debug's own log-failure diagnostics previously relied on the automatic setter. Paired [Goode PR207](https://github.com/timwhitez/goode/pull/207) migrates logEntry/warnf (including streaming/error paths) and two old setter-dependent headless fixtures to the context binding. Both PRs must be independently reviewed and ready before merging SDK then Goode. Issue133 stays open until paired main validation; do not treat SDK alone as full delivery. No pin-only workflow; validate actual selected SDK sources.

## Validation

Author and independent exact6a538ff full/vet/LLM-provider-Agent-compaction race and focusedrace100 passed. Eight shared Agent invocations get exactly their own warning in buffered/streaming paths, no construction setter calls; all three concrete clients keep direct fallback while concurrent contexts isolate warnings; compaction uses its own Service scope; context cancellation/deadline/nil masking preserved. Shared-client construction plus real Invoke asserts configured80/child0 for all three clients. Independent six mutations caught restoredNewsetter, missingAgent/Servicecontext and eachprovider ignoringbinding; restoredclean. SDKreviewer APPROVE. Paired Goode finalreview still required before merge.

No logging-content expansion, model/tool/Prompt behavior change, cache mapping change or benchmark speedup claim. This fixes construction race and invocation ownership, not arbitrary concurrent mutation of public client config or uncooperative callback behavior.

### DIFF
--- a/agent_docs/providers.md
+++ b/agent_docs/providers.md
@@ -23,10 +23,13 @@ stream normalization, and response metadata behavior.
   and [tool-definition API](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-use-with-prompt-caching).
   Legacy `Message.Cache` behavior is unchanged; do not use the new field as a
   cache-control guarantee. No Agent Config or host plan generator is added.
-  `CachePlan` now includes a private binding. Responses diagnostics use its
-  construction-time `Warningf` (or the default logger), not automatic shared
-  `WarningSinkSetter` mutation during child Agent creation. Configure it before
-  concurrent use. External unkeyed literals need updating;
+  `CachePlan` includes a private binding. Agent/Compaction invocations carry
+  `llm.WithWarningSink` in context; built-in clients capture it in call-local
+  state. Agent.New never mutates a shared model's `WarningSinkSetter`. Direct
+  clients retain their construction-time Warningf/default logger; configure
+  fields/setters before concurrent use. Custom providers must read
+  `llm.WarningSink(ctx, fallback)` to honor invocation routing. External unkeyed
+  literals affected by the earlier CachePlan binding still need updating;
   JSON cannot restore a valid binding. Message/Completion layouts and session
   JSON remain unchanged. No content fingerprints or Git-hash gates are added.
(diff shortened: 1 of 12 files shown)


timwhitez:
Independent SDK reviewer pin194_review APPROVE exact6a538fffd7b4b7b14ff1e9024d5601af6a9b2ea5: full/vet/build/LLM-provider-Agent-Compaction race/focused100 passed. Six mutations detected sharedSetter restoration, missingAgent/Service bindings, and eachprovider ignoringcontext; restoredclean. Companion Goode207 final6ed9560 independently APPROVED with full/vet/sixpkg race/strictactualSDKbuild, originalprobes and two migratedheadlessfixtures race10; no assertion weakening. Both heads ready; merge SDK then Goode and close133 only after dual-main verification. No pin-only workflow.

timwhitez:
Paired main verification complete with SDK08bce6d / Goode dc0e459: full/vet/LLM-provider-Agent-Compaction race, Goode sixpkg race/strictbuild and original probes race100 all pass. SDK focused ownership100 plus Goode headless order/drain/debug fixtures10 pass. Issue133 closed only after this evidence. Logs /tmp/sdk134-main-*.log, /tmp/goode207-main-*.log, /tmp/warning133-main-original-probes.log.

Issue #86, opened 2026-09-01
[agent/compaction] 统一 CompactionController：触发、抑制与 bounded overflow recovery


timwhitez:
## 当前状态：自动决策 Shadow 与安全 publication 基础已完成

基线 SDK `7994605` / Goode `c55126a`。此 Issue 是控制决策收敛，不是替换现有 Compaction Engine。

## 已完成，不重复建设

- #98 Characterization：usage threshold/tier、todo/retry 优先级、placeholder pressure、cooldown/inflight/pending、hard-usage overflow bypass、Provider 400 terminal。见 `TestCompactionLegacyDecisionParity`、`TestProviderRejectedContextOverflowIsTerminal`。
- #102：`compaction_decision.go` 与 `checkAndCompactWithGrowth` 的 private automatic-admission `{run, trigger, targetWatermark}` Shadow；admission/cooldown 只采样一次，现有决策仍驱动行为，日志仅安全枚举/布尔。见 `compaction_decision_test.go`。
- 自动与 `CompactPipelineNow` 已复用 canonical pipeline；`CompactNow` 明确请求 manual/summary/force-summary。manual force-summary、local no-provider、checkpoint failure、runtime replacement barrier、pending publication 均有测试。
- #120/#122/#123 与 Goode #198/#199/#202：checked active history、public manual query admission、`CommitCompactionHistory(ctx, expected, candidate, result)` source-checked publication。Goode 生产 fallback 已不再采用 checkpoint→外部 ReplaceHistory 两步提交；确认后的 publication 即使随后取消也完成，未知提交结果不虚报未执行。Goode #201 另修复 preflight/fallback 错误分类与停止边界。

`CommitCompactionHistory` 复用现有 admission/checkpoint/ledger，content equality 不是 Session/runtime revision。Trim 候选捕获早于 lease；外部 store writer 与独立 legacy checkpoint-only API 不受这项事务保护。`compactionGeneration` 计成功 apply，不是 config identity。历史 Review、mutation、微基准记录在对应 PR，不再作为新的实施阶段。

## 真实缺口

1. 自动、manual 与 host preflight 尚未共享唯一 Decision/policy authority；触发、force、阈值和fallback行为存在真实差异。
2. 尚无 provider-neutral typed context-overflow recovery 和按 accepted real-user epoch 的有界重准入预算；现有 `compactSyncOverflow` 是 usage-driven local recovery，不等同于任意 Provider 400/422 恢复。
[...]

PR #141, opened 2026-09-12
test(agent): characterize compaction summary attempt budgets

timwhitez:
Related to #86; this multi-phase issue remains open. Test-only characterization through real automatic/public entry paths, not only plan return values: configured full retry budget, armed cooldown single attempt, failure-streak zero summary attempts, cancellation zero attempts; public CompactNow and explicit allow/deny/force retain their own policy under automatic failure suppression. Reuse the existing fake; do not duplicate the existing default-two boundary test. Assert actual summary model Invoke counts and unchanged history on failure/no-op. No production Controller/API, threshold, provider payload, prompt or persistence changes. Focused tests passed; full/vet/agent+compaction race/build and independent exact-head Review/mutations in progress. No claim of real provider HTTP/remote cache performance.

### DIFF
--- a/sdk/agent/compaction_summary_budget_test.go
+++ b/sdk/agent/compaction_summary_budget_test.go
@@ -1,0 +1,108 @@
+package agent
+
+import (
+	"context"
+	"errors"
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/timwhitez/agent-sdk-golang/sdk/agent/compaction"
+	"github.com/timwhitez/agent-sdk-golang/sdk/llm"
+)
+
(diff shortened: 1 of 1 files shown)


timwhitez:
Final exact-head independent Review APPROVE 7493506912b87292a778de1a1dbaa9efabe9b80c. Independent full/vet/build, Agent/Compaction race and focused race x100 passed; three mutations caught budget/policy regressions and were restored. Author full/vet/Agent+Compaction race/build terminalexit0 and restored race x20 passed. Scope is explicit attempt-budget characterization through real entry paths; failure state is a fixture and cancellation is pre-cancel only, not a new Controller/whole Query implementation. Issue86 remains open for equivalent policy consolidation and later enforcement.

